PR #1, opened 2022-03-24
Add first tools for Docker sysext images

pothos:
The tools address Flatcar's use case of running a custom Docker version
which is currently done with custom binaries under /opt/bin or Torcx.
The Flatcar docs will point to these tools
(https://github.com/flatcar-linux/flatcar-docs/pull/232).


README.md
```diff
@@ -0,0 +1,94 @@
...
+While systemd-sysext images are not really meant to also include the systemd service, Flatcar ships `ensure-sysext.service` as workaround to automatically load the image's services.
+It reloads the unit files from disk and reevaluates `multi-user.target`, `sockets.target`, and `timers.target`, making sure your enabled systemd units run.
+This service is bound to `systemd-sysext.service` which activates the sysext images on boot.
+At runtime executing `systemctl restart systemd-sysext` will reload the sysext images and start the services.
+A manual `systemd-sysext refresh` is not recommended.
+
+The compatibility mechanism of sysext images requires a metadata file in the image under `usr/lib/extension-release.d/extension-release.NAME`.
+It has to contain a matching OS `ID`, and either a matching `VERSION` or `SYSEXT_LEVEL`.
+Since the rapid release cycle and automatic updates of Flatcar Container Linux make it hard to rely on particular OS libraries by specifying a dependency of the sysext image to the OS version, it is not recommended to match by `VERSION`.
```
tormath1, line 19:
```suggestion
Since the rapid release cycle and automatic updates of Flatcar Container Linux makes it hard to rely on particular OS libraries by specifying a dependency of the sysext image to the OS version, it is not recommended to match by `VERSION`.
```

README.md
```diff
@@ -0,0 +1,94 @@
...
+Instead, Flatcar defined the `SYSEXT_LEVEL` value `1.0` to match for.
+The sysext image should only include static binaries.
+
+Inside the image, binaries should be placed under `usr/bin/` and systemd units under `usr/lib/systemd/system/`.
+To enable systemd units symlinks should be included in the image itself the same way as systemd would normally generate them when enabling the units, e.g., `sockets.target.wants/my.socket` → `../my.socket`.
```
tormath1, line 24:
```suggestion
To enable systemd units, symlinks should be included in the image itself in the same way as systemd would normally generate them when enabling the units, e.g., `sockets.target.wants/my.socket` → `../my.socket`.
```

README.md
pothos: I think "make" is correct for the plural "A and B"?

tormath1: You're right, I was only considering the "rapid release cycle". :)

tormath1 (approved):
LGTM :)

convert_torcx_image.sh
```diff
@@ -0,0 +1,39 @@
...
+mkdir -p "${SYSEXTNAME}"/usr
+if [ -e "${SYSEXTNAME}"/bin ]; then
+  mv "${SYSEXTNAME}"/bin "${SYSEXTNAME}"/usr/
+fi
+if [ -e "${SYSEXTNAME}"/lib ]; then
+  mv "${SYSEXTNAME}"/lib "${SYSEXTNAME}"/usr/
+fi
+for FILE in "${SYSEXTNAME}"/usr/lib/systemd/system/*.service "${SYSEXTNAME}"/usr/lib/systemd/system/*.socket; do
+  if [ -e "${FILE}" ]; then
+    sed -i -e 's,${TORCX_BINDIR},/usr/bin,g' -e 's,${TORCX_UNPACKDIR},/,g' -e 's,${TORCX_IMAGEDIR},,g' \
```
jepio, line 29: What do you think of overwriting the environment with `Environment=TORCX_BINDIR=/usr/bin`, etc. instead? Or overwriting `EnvironmentFile` (is that possible?) to a file contained within the sysext. That way drop-ins continue working? Or is this carrying too much legacy with us.

pothos: The problem with the drop-ins is that they stop working as soon as one would switch to the inbuilt sysext image. We could start to encode these variables there, too, though…?

pothos: Ok, updated it now to set the environment

PR #6, opened 2022-11-11
create_docker_sysext.sh: Allow to limit sysext to Docker or containerd

pothos:
- create_docker_sysext.sh: Allow to limit sysext to Docker or containerd
    
    With Kubernetes users may want to only ship containerd instead of also
    Docker. At the same time it may be handy to layer Docker on top if
    needed.
    Introduce arguments that allow to limit the build to one of Docker or
    containerd.
- create_docker_sysext.sh: Enable containerd.service by default
    
    This is also done in Flatcar for Kubernetes 1.24.
- create_docker_sysext.sh: Add workaround for missing squashfs tool
    
    On Flatcar we don't have mksquashfs but still want to create sysext
    images for testing.
    Allow to use btrfs or ext4 as alternative format but don't recommend
    it.

## How to use/Testing done

Serve the images locally after building with `ONLY_DOCKER=1 ./create_docker_sysext.sh 20.10.21 docker` and `ONLY_CONTAINERD=1 ./create_docker_sysext.sh 20.10.21 containerd`. Then boot up with this Ignition config:
```
{
  "ignition": {
    "version": "3.3.0"
  },
  "passwd": {
    "users": [
      {
        "name": "core",
        "sshAuthorizedKeys": [
          "ssh-rsa AAA…"
        ]
      }
    ]
  },
  "storage": {
    "files": [
      {
        "path": "/etc/extensions/docker.raw",
        "contents": {
          "source": "http://192.168.1.…:8000/docker.raw"
        },
        "mode": 420
      },
      {
       	"path": "/etc/extensions/containerd.raw",
        "contents": {
          "source": "http://192.168.1.…:8000/containerd.raw"
        },
	"mode": 420
      },
      {
        "path": "/etc/systemd/system-generators/torcx-generator"
      }
    ],
    "directories": [
      {
        "path": "/etc/extensions/docker-flatcar"
      },
      {
        "path": "/etc/extensions/containerd-flatcar"
      }
    ]
  }
}

```

Check with
```
core@localhost ~ $ systemd-sysext status
HIERARCHY EXTENSIONS SINCE                      
/opt      none       -                          
/usr      containerd Fri 2022-11-11 14:49:42 UTC
          docker     
```
Run `docker` or `ctr` and see that they are static binaries.
Check that unit files under `/usr` are used with `systemctl status docker containerd`.

### DIFF
--- a/create_docker_sysext.sh
+++ b/create_docker_sysext.sh
@@ -3,6 +3,8 @@ set -euo pipefail
 
 ARCH="${ARCH-x86_64}"
 OS="${OS-flatcar}"
+ONLY_CONTAINERD="${ONLY_CONTAINERD:-0}"
+ONLY_DOCKER="${ONLY_DOCKER:-0}"
 
 if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
   echo "Usage: $0 VERSION SYSEXTNAME"
@@ -10,10 +12,17 @@ if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
   echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
   echo "All files in the sysext image will be owned by root."
   echo "The necessary systemd services will be created by this script, by default only docker.socket will be enabled."
+  echo "To only package containerd without Docker, pass ONLY_CONTAINERD=1 as environment variable (current value is '${ONLY_CONTAINERD}')."
+  echo "To only package Docker without containerd and runc, pass ONLY_DOCKER=1 as environment variable (current value is '${ONLY_DOCKER}')."
   echo "To use arm64 pass 'ARCH=aarch64' as environment variable (current value is '${ARCH}')."
   echo "To build for another OS than Flatcar, pass 'OS=myosid' as environment variable (current value is '${OS}'), e.g., 'fedora' as found in 'ID' under '/etc/os-release'."
   echo "The '/etc/os-release' file of your OS has to include 'SYSEXT_LEVEL=1.0' as done in Flatcar."
   echo
+  exit 1
+fi
+
+if [ "${ONLY_CONTAINERD}" = 1 ] && [ "${ONLY_DOCKER}" = 1 ]; then
+  echo "Cannot set both ONLY_CONTAINERD and ONLY_DOCKER" >&2
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


create_docker_sysext.sh
```diff
@@ -3,25 +3,37 @@ set -euo pipefail
...
   echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
   echo "All files in the sysext image will be owned by root."
   echo "The necessary systemd services will be created by this script, by default only docker.socket will be enabled."
+  echo "To only package containerd without Docker, pass ONLY_CONTAINERD=1 as environment variable (current value is '${ONLY_CONTAINERD}')."
+  echo "To only package Docker without containerd and runc, pass ONLY_DOCKER=1 as environment variable (current value is '${ONLY_DOCKER}')."
   echo "To use arm64 pass 'ARCH=aarch64' as environment variable (current value is '${ARCH}')."
   echo "To build for another OS than Flatcar, pass 'OS=myosid' as environment variable (current value is '${OS}'), e.g., 'fedora' as found in 'ID' under '/etc/os-release'."
   echo "The '/etc/os-release' file of your OS has to include 'SYSEXT_LEVEL=1.0' as done in Flatcar."
+  echo "If you miss the mksquashfs tool you can pass FORMAT=btrfs or FORMAT=ext4 as environment variable but the files won't be owned by root."
```
krnowak, line 21: To miss a tool sounds weird (you usually miss someone close when you don't see them for a long time).

```suggestion
  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs or FORMAT=ext4 as environment variable but the files won't be owned by root."
```

create_docker_sysext.sh
```diff
@@ -3,25 +3,37 @@ set -euo pipefail
 
 ARCH="${ARCH-x86_64}"
 OS="${OS-flatcar}"
+ONLY_CONTAINERD="${ONLY_CONTAINERD-0}"
+ONLY_DOCKER="${ONLY_DOCKER-0}"
+FORMAT="${FORMAT-squashfs}"
```
krnowak, line 8: How about being stricter and validating this variable to be either "squashfs" or "ext4" or "btrfs" early?

create_docker_sysext.sh
```diff
@@ -3,25 +3,37 @@ set -euo pipefail
 
 ARCH="${ARCH-x86_64}"
 OS="${OS-flatcar}"
+ONLY_CONTAINERD="${ONLY_CONTAINERD-0}"
+ONLY_DOCKER="${ONLY_DOCKER-0}"
```
krnowak, line 7: I was wondering if it would make sense for possible values of these variables to be "empty" and "non-empty" instead of "1" and "anything else". You could then test them with `[[ -z … ]]` or `[[ -n … ]]` and avoid the jarring ` = 1` or `!= 1` (likely ought to be `-eq 1` and `-ne 1`). :)

But this is up to you. Not a blocker, certainly.

pothos: Thanks, if we use `:-0` the empty case will be handled to default to `0`.

create_docker_sysext.sh
pothos: :+1:

krnowak (approved):

PR #7, opened 2023-01-03
Add wasmtime sysext helper

pothos:
Even though wasmtime is a single binary (which seems to link to an old enough glibc), it's a good showcase for sysext images because using a Docker container to run wasmtime would add unnecessary overhead when spawning many instances.

## How to use/Testing done

```
./create_wasmtime_sysext.sh 4.0.0 flatcar-wasmtime
scp -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -P 2222 flatcar-wasmtime.raw  core@127.0.0.1:
ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p 2222 core@127.0.0.1
# then
sudo mkdir -p /etc/extensions/
sudo mv flatcar-wasmtime.raw /etc/extensions/
sudo systemctl restart systemd-sysext
```

### DIFF
--- a/create_wasmtime_sysext.sh
+++ b/create_wasmtime_sysext.sh
@@ -13,13 +13,13 @@ if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
   echo "To use arm64 pass 'ARCH=aarch64' as environment variable (current value is '${ARCH}')."
   echo "To build for another OS than Flatcar, pass 'OS=myosid' as environment variable (current value is '${OS}'), e.g., 'fedora' as found in 'ID' under '/etc/os-release'."
   echo "The '/etc/os-release' file of your OS has to include 'SYSEXT_LEVEL=1.0' as done in Flatcar."
-  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs or FORMAT=ext4 as environment variable but the files won't be owned by root."
+  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2 as environment variable but the files won't be owned by root."
   echo
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


create_wasmtime_sysext.sh
```diff
@@ -0,0 +1,54 @@
...
+rm -f "${SYSEXTNAME}".raw
+if [ "${FORMAT}" = "btrfs" ]; then
+  # Note: We didn't chown to root:root, meaning that the file ownership is left as is
+  mkfs.btrfs --mixed -m single -d single --shrink --rootdir "${SYSEXTNAME}" "${SYSEXTNAME}".raw
+  # This is for testing purposes and makes not much sense to use because --rootdir doesn't allow to enable compression
+elif [ "${FORMAT}" = "ext4" ]; then
+  # Assuming that 1 GB is enough
+  truncate -s 1G "${SYSEXTNAME}".raw
+  # Note: We didn't chown to root:root, meaning that the file ownership is left as is
+  mkfs.ext4 -E root_owner=0:0 -d "${SYSEXTNAME}" "${SYSEXTNAME}".raw
```
jepio, line 49: How about formatting as ext2, since the image will be mounted ro (=doesn't benefit from a journal, and the journal takes up space).
Also let's add this at the end:
```
```
ext2+resize results in 22M image file vs 70M for ext4+resize.

jepio (approved):
Looks good

create_wasmtime_sysext.sh
pothos: Good idea, ext4 was mainly a workaround for use in kola because Flatcar doesn't have the squashfs tools (and the btrfs tools are outdated and miss fixes).

pothos: Pushed the change

PR #8, opened 2023-03-20
Use Upholds= to ensure services run after daemon-reload

pothos:
- Use Upholds= to ensure services run after daemon-reload
    
    When a systemd-sysext image is loaded, normally a daemon-reload is done
    but the targets are not evaluated, meaning that enabled services don't
    start. In Flatcar we ship a workaround for that by restarting a certain
    list of targets but this is not upstream behavior. Currently upstream
    also doesn't do the daemon-reload but we can file an upstream PR for
    that.
- gitignore: Ignore any local tar balls
- Handle tar files with colons in their name
    
    The use of colons in a tar filename argument will invoke rsh/ssh to
    fetch it from the domain interpreted from the part before the colon.
    This is not what we want and we have to opt out of this behavior.

## Testing done

Tested the custom Docker sysext built from release binaries


README.md
```diff
@@ -9,19 +9,27 @@ The current focus is on Docker and containerd, contributions are welcome for oth
...
 While systemd-sysext images are not really meant to also include the systemd service, Flatcar ships `ensure-sysext.service` as workaround to automatically load the image's services.
-It reloads the unit files from disk and reevaluates `multi-user.target`, `sockets.target`, and `timers.target`, making sure your enabled systemd units run.
-This service is bound to `systemd-sysext.service` which activates the sysext images on boot.
-At runtime executing `systemctl restart systemd-sysext` will reload the sysext images and start the services.
+This helper service is bound to `systemd-sysext.service` which activates the sysext images on boot.
+Currently it reloads the unit files from disk and reevaluates `multi-user.target`, `sockets.target`, and `timers.target`, making sure your enabled systemd units run.
+In the future `systemd-sysext` will only reload the unit files when this is upstream behavior (the current upstream behavior is to do nothing and leave it to the user).
+That means you will have to use `Upholds=` drop-ins for the target units to start your units.
```
krnowak, line 15:
```suggestion
That means you will need to use `Upholds=` drop-ins for the target units to start your units.
```

README.md
```diff
@@ -9,19 +9,27 @@ The current focus is on Docker and containerd, contributions are welcome for oth
...
+At runtime executing `systemctl restart systemd-sysext ensure-sysext` will reload the sysext images and start the services.
 A manual `systemd-sysext refresh` is not recommended.
 
 The compatibility mechanism of sysext images requires a metadata file in the image under `usr/lib/extension-release.d/extension-release.NAME`.
-It has to contain a matching OS `ID`, and either a matching `VERSION` or `SYSEXT_LEVEL`.
-Since the rapid release cycle and automatic updates of Flatcar Container Linux make it hard to rely on particular OS libraries by specifying a dependency of the sysext image to the OS version, it is not recommended to match by `VERSION`.
+It has to contain a matching OS `ID`, and either a matching `VERSION_ID` or `SYSEXT_LEVEL`.
```
krnowak, line 20:
```suggestion
It needs to contain a matching OS `ID`, and either a matching `VERSION_ID` or `SYSEXT_LEVEL`.
```

README.md
```diff
@@ -9,19 +9,27 @@ The current focus is on Docker and containerd, contributions are welcome for oth
...
+Since the rapid release cycle and automatic updates of Flatcar Container Linux make it hard to rely on particular OS libraries by specifying a dependency of the sysext image to the OS version, it is not recommended to match by `VERSION_ID`.
 Instead, Flatcar defined the `SYSEXT_LEVEL` value `1.0` to match for.
 The sysext image should only include static binaries.
 
 Inside the image, binaries should be placed under `usr/bin/` and systemd units under `usr/lib/systemd/system/`.
-To enable systemd units, symlinks should be included in the image itself in the same way as systemd would normally generate them when enabling the units, e.g., `sockets.target.wants/my.socket` → `../my.socket`.
+To enable systemd units, while it is currently supported to place symlinks in the image itself in the same way as systemd would normally generate them when enabling the units, e.g., `sockets.target.wants/my.socket` → `../my.socket`, this is not recommended.
```
krnowak, line 26:
```suggestion
While placing symlinks in the image itself to enable the units in the same way as systemd would normally do (by generating symlinks like `sockets.target.wants/my.socket` → `../my.socket`) is still currently supported, this is not a recommended practice.
```

README.md
```diff
@@ -9,19 +9,27 @@ The current focus is on Docker and containerd, contributions are welcome for oth
...
+The recommended way is to ship drop-ins for the target units that start your unit, e.g., `/usr/lib/systemd/system/sockets.target.d/10-docker-socket.conf` would contain (similar for `multi-user.target` and a `.service` unit):
```
krnowak, line 27: The recommended way is to ship drop-ins for the target units that start your unit. The drop-in file should use the `Upholds` clause. For example, a drop-in for `docker.socket` that is started by `sockets.target` would need to be placed in `/usr/lib/systemd/system/sockets.target.d/10-docker-socket.conf` with the following contents:

README.md
```diff
@@ -9,19 +9,27 @@ The current focus is on Docker and containerd, contributions are welcome for oth
...
+
+```
+[Unit]
+Upholds=docker.socket
+```
```
krnowak, line 33: Here I'd add something like:

```
This can be done also for services, so for `docker.service` started by `multi-user.target`, the drop-in would reside in `/usr/lib/systemd/system/multi-user.target.d/10-docker-service.conf` and it would have `Upholds=docker.service` line.
```

convert_torcx_image.sh
```diff
@@ -32,6 +32,12 @@ for FILE in "${SYSEXTNAME}"/usr/lib/systemd/system/*.service "${SYSEXTNAME}"/usr
       -e 's,After=torcx.target,,g' -e 's,Requires=torcx.target,,g' "${FILE}"
   fi
 done
+for ENTRY in "${SYSEXTNAME}/usr/lib/systemd/system/"*.wants/*; do
+  UNIT=$(echo "${ENTRY}" | rev | cut -d / -f 1 |rev)
+  TARGET=$(echo "${ENTRY}" | rev | cut -d / -f 2 |rev | sed 's/.wants$//g')
```
krnowak, line 37: This could be basename of dirname, so `TARGET=$(basename "$(dirname "${ENTRY}")" | sed 's/\.wants$//'), but it's up to you - it's not that much simpler than what you wrote. So suggesting just some minor things:

```suggestion
  TARGET=$(echo "${ENTRY}" | rev | cut -d / -f 2 | rev | sed 's/.wants$//')
```

convert_torcx_image.sh
```diff
@@ -32,6 +32,12 @@ for FILE in "${SYSEXTNAME}"/usr/lib/systemd/system/*.service "${SYSEXTNAME}"/usr
       -e 's,After=torcx.target,,g' -e 's,Requires=torcx.target,,g' "${FILE}"
   fi
 done
+for ENTRY in "${SYSEXTNAME}/usr/lib/systemd/system/"*.wants/*; do
+  UNIT=$(echo "${ENTRY}" | rev | cut -d / -f 1 |rev)
```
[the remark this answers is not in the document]
pothos: Yes, sounds like basename is what I forgot to use :)

README.md
pothos: Thanks

README.md
pothos: Thanks

krnowak (approved):

PR #9, opened 2023-03-24
Move sysext image creation to shared helper

pothos:
The code to package the folder in filesystem image was shared. To ease maintenance and adding new features like systemd-repart for GPT images with dm-verity, move it to an own helper. While doing so we can also support the new properties that we added in systemd 252.

## How to use

You can use the `bake.sh` script also standalone to make an image of a folder you prepared.

## Testing done

Tested wasmtime, docker release, and torcx conversion sysext images for amd64 and arm64, checked on Alpha that the ARCHITECTURE field and OS=_any works.

### DIFF
--- a/bake.sh
+++ b/bake.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+set -euo pipefail
+
+OS="${OS-flatcar}"
+FORMAT="${FORMAT:-squashfs}"
+ARCH="${ARCH-}"
+
+# This script is to be called as helper by other scripts but can also be used standalone
+if [ $# -lt 1 ]; then
+  echo "Usage: $0 SYSEXTNAME"
+  echo "The script will make a SYSEXTNAME.raw image of the folder SYSEXTNAME, and create an os-release file in it, run with --help for the list of supported environment variables."
+  exit 1
+elif [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "If ARCH is specified as environment variable the sysext image will be required to run on the given architecture."
+  echo "To build for another OS than Flatcar, pass 'OS=myosid' as environment variable (current value is '${OS}'), e.g., 'fedora' as found in 'ID' under '/etc/os-release', or pass 'OS=_any' for any OS."
+  echo "The '/etc/os-release' file of your OS has to include 'SYSEXT_LEVEL=1.0' as done in Flatcar (not needed for 'OS=_any')."
+  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2 as environment variable (current value is '${FORMAT}') but the script won't change the ownership of the files in the SYSEXTNAME directory, so make sure that they are owned by root before creating the sysext image to avoid any problems."
+  echo
+  exit 1
+fi
+
+SYSEXTNAME="$1"
+
+if [ "${FORMAT}" != "squashfs" ] && [ "${FORMAT}" != "btrfs" ] && [ "${FORMAT}" != "ext4" ] && [ "${FORMAT}" != "ext2" ]; then
+  echo "Expected FORMAT=squashfs, FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2, got '${FORMAT}'" >&2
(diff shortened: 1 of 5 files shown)


pothos:
Updated the README, too

bake.sh
```diff
@@ -0,0 +1,59 @@
...
+
+if [ "${FORMAT}" != "squashfs" ] && [ "${FORMAT}" != "btrfs" ] && [ "${FORMAT}" != "ext4" ] && [ "${FORMAT}" != "ext2" ]; then
+  echo "Expected FORMAT=squashfs, FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2, got '${FORMAT}'" >&2
+  exit 1
+fi
+
+# Map to valid values for https://www.freedesktop.org/software/systemd/man/os-release.html#ARCHITECTURE=
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86_64" ]; then
+  ARCH="x86-64"
+elif [ "${ARCH}" = "aarch64" ]; then
```
krnowak, line 31: Let's support arm64 as an arch too.

```suggestion
elif [ "${ARCH}" = "arm64" ] || [ "${ARCH}" = "aarch64" ]; then
```

bake.sh
```diff
@@ -0,0 +1,59 @@
...
+mkdir -p "${SYSEXTNAME}/usr/lib/extension-release.d"
+{
+  echo "ID=${OS}"
+  if [ "${OS}" != "_any" ]; then
+    echo "SYSEXT_LEVEL=1.0"
+  fi
+  if [ "${ARCH}" != "" ]; then
+    echo "ARCHITECTURE=${ARCH}"
+  fi
+} > "${SYSEXTNAME}/usr/lib/extension-release.d/extension-release.${SYSEXTNAME}"
```
krnowak, line 44: No support for setting `VERSION_ID`? I suppose it's not necessary for the script for now.

bake.sh
```diff
@@ -0,0 +1,59 @@
...
+
+# This script is to be called as helper by other scripts but can also be used standalone
+if [ $# -lt 1 ]; then
+  echo "Usage: $0 SYSEXTNAME"
+  echo "The script will make a SYSEXTNAME.raw image of the folder SYSEXTNAME, and create an os-release file in it, run with --help for the list of supported environment variables."
+elif [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "If ARCH is specified as environment variable the sysext image will be required to run on the given architecture."
+  echo "To build for another OS than Flatcar, pass 'OS=myosid' as environment variable (current value is '${OS}'), e.g., 'fedora' as found in 'ID' under '/etc/os-release', or pass 'OS=_any' for any OS."
+  echo "The '/etc/os-release' file of your OS has to include 'SYSEXT_LEVEL=1.0' as done in Flatcar (not needed for 'OS=_any')."
+  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2 as environment variable (current value is '${FORMAT}') but the files won't be owned by root."
```
krnowak, line 16:
```suggestion
  echo "If the mksquashfs tool is missing you can pass FORMAT=btrfs, FORMAT=ext4, or FORMAT=ext2 as environment variable (current value is '${FORMAT}') but the script won't change the ownership of the files in SYSEXTNAME directory, so make sure that they are owned by root before creating the sysext image."
```

bake.sh
pothos: `arm64` is already in the valid value and doesn't need to be converted

bake.sh
pothos: Yes, I don't see this use case and would assume that we have our own commands in Flatcar for the version-specific builds

bake.sh
pothos: Thanks, the sysext is mounted read-only, but it still makes sense to hint people into setting this up correctly to avoid any problems when a mutable overlay exists.

bake.sh
krnowak: Oh, right. Nevermind then. :)

krnowak (approved):

PR #14, opened 2023-07-20
bakery: add kubernetes sysext helper

tormath1:
## How to use

`./create_kubernetes_sysext.sh 1.27.3 kubernetes` then

```yaml
variant: flatcar
version: 1.0.0
storage:
  files:
    - path: /opt/extensions/kubernetes.raw
      mode: 0644
      contents:
        source: http://.../kubernetes.raw
  links:
    - path: /etc/extensions/kubernetes.raw
      hard: false
      target: /opt/extensions/kubernetes.raw
```

## Tests

* Tested on QEMU (simply deploy a control plane)

* Tested on OpenStack CAPI (CAPO) with the following template diff:
```diff
+            storage:
+              files:
+                - path: /opt/extensions/kubernetes.raw
+                  mode: 0644
+                  contents:
+                    remote:
+                      url: https://storage.googleapis.com/tormath1-flatcar/kubernetes.raw
+              links:
+                - filesystem: root
+                  path: /etc/extensions/kubernetes.raw
+                  hard: false
+                  target: /opt/extensions/kubernetes.raw
             systemd:
               units:
               - name: coreos-metadata-sshkeys@.service
@@ -106,6 +118,18 @@ spec:
     ignition:
       containerLinuxConfig:
         additionalConfig: |
+          storage:
+            files:
+              - path: /opt/extensions/kubernetes.raw
+                mode: 0644
+                contents:
+                  remote:
+                    url: https://storage.googleapis.com/tormath1-flatcar/kubernetes.raw
+            links:
+              - filesystem: root
+                path: /etc/extensions/kubernetes.raw
+                hard: false
+                target: /opt/extensions/kubernetes.raw
           systemd:
             units:
             - name: coreos-metadata-sshkeys@.service
```

On the node:
```
$ systemd-sysext status
HIERARCHY EXTENSIONS SINCE
/opt      none       -
/usr      kubernetes Mon 2023-07-24 16:01:05 UTC
```

Nodes:
```
$  kubectl --kubeconfig=./${CLUSTER_NAME}.kubeconfig get nodes
NAME                                  STATUS   ROLES           AGE     VERSION
capi-quickstart-control-plane-c5lkk   Ready    control-plane   7m34s   v1.27.3
capi-quickstart-md-0-hdsf5            Ready    <none>          6m55s   v1.27.3
capi-quickstart-md-0-kjj9m            Ready    <none>          6m56s   v1.27.3
capi-quickstart-md-0-qdbcn            Ready    <none>          6m56s   v1.27.3
$ kubectl get kubeadmcontrolplane
NAME                            CLUSTER           INITIALIZED   API SERVER AVAILABLE   REPLICAS   READY   UPDATED   UNAVAILABLE   AGE     VERSION
capi-quickstart-control-plane   capi-quickstart   true          true                   1          1       1         0             8m53s   v1.27.2
```

### DIFF
--- a/create_kubernetes_sysext.sh
+++ b/create_kubernetes_sysext.sh
@@ -1,0 +1,96 @@
+#!/bin/bash
+set -euo pipefail
+
+export ARCH="${ARCH-amd64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME [CNI_VERSION]"
+  echo "The script will download the Kubernetes release binaries (e.g., for v1.27.3) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
+  echo "All files in the sysext image will be owned by root."
+  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
+  echo "CNI version current value is 'latest'"
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+CNI_VERSION="${3-latest}"
+if [ "${ARCH}" = aarch64 ]; then
+  ARCH=arm64
+fi
+rm -f kubectl kubeadm kubelet
+
+# install kubernetes binaries.
+curl -o kubectl -fsSL "https://dl.k8s.io/${VERSION}/bin/linux/${ARCH}/kubectl"
+curl -o kubeadm -fsSL "https://dl.k8s.io/${VERSION}/bin/linux/${ARCH}/kubeadm"
+curl -o kubelet -fsSL "https://dl.k8s.io/${VERSION}/bin/linux/${ARCH}/kubelet"
+
+rm -rf "${SYSEXTNAME}"
+mkdir -p "${SYSEXTNAME}"/usr/bin
+mv kubectl kubeadm kubelet "${SYSEXTNAME}"/usr/bin
+
+chmod +x "${SYSEXTNAME}"/usr/bin/{kubectl,kubeadm,kubelet}
+
+# setup kubelet service.
+mkdir -p "${SYSEXTNAME}/usr/lib/systemd/system"
+cat > "${SYSEXTNAME}/usr/lib/systemd/system/kubelet.service" <<-'EOF'
+[Unit]
+Description=kubelet: The Kubernetes Node Agent
+Documentation=https://kubernetes.io/docs/home/
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+ExecStart=/usr/bin/kubelet
+Restart=always
+StartLimitInterval=0
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+mkdir -p "${SYSEXTNAME}/usr/lib/systemd/system/kubelet.service.d"
+cat > "${SYSEXTNAME}/usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf" <<-'EOF'
+# Note: This dropin only works with kubeadm and kubelet v1.11+
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--bootstrap-kubeconfig=/etc/kubernetes/bootstrap-kubelet.conf --kubeconfig=/etc/kubernetes/kubelet.conf"
+Environment="KUBELET_CONFIG_ARGS=--config=/var/lib/kubelet/config.yaml"
+# This is a file that "kubeadm init" and "kubeadm join" generates at runtime, populating the KUBELET_KUBEADM_ARGS variable dynamically
+EnvironmentFile=-/var/lib/kubelet/kubeadm-flags.env
+# This is a file that the user can use for overrides of the kubelet args as a last resort. Preferably, the user should use
+# the .NodeRegistration.KubeletExtraArgs object in the configuration files instead. KUBELET_EXTRA_ARGS should be sourced from this file.
+EnvironmentFile=-/etc/sysconfig/kubelet
+ExecStart=
+ExecStartPre=/usr/bin/mkdir -p /opt/libexec /opt/libexec.work
+ExecStartPre=/usr/bin/cp -r /usr/local/bin/cni/ /opt/bin/cni
+ExecStartPre=/usr/bin/cp /usr/local/share/kubernetes-version /etc/kubernetes-version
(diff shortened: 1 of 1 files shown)


create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,99 @@
...
+
+export ARCH="${ARCH-amd64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download the wasmtime release tar ball (e.g., for 4.0.0) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
+  echo "All files in the sysext image will be owned by root."
+  echo "To use arm64 pass 'ARCH=aarch64' as environment variable (current value is '${ARCH}')."
```
pothos, line 12:
```suggestion
  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
```

pothos: The K8s URLs are expecting this

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,99 @@
...
+WantedBy=multi-user.target
+EOF
+
+mkdir -p "${SYSEXTNAME}/usr/lib/systemd/system/multi-user.target.d"
+{ echo "[Unit]"; echo "Upholds=kubelet.service"; } > "${SYSEXTNAME}/usr/lib/systemd/system/multi-user.target.d/10-kubelet-service.conf"
+{ echo "[Unit]"; echo "Upholds=usr-libexec.mount"; } > "${SYSEXTNAME}/usr/lib/systemd/system/multi-user.target.d/10-usr-libexec.conf"
+
+# install CNI.
+# TODO: use CNI version / arch variable
+curl -o cni.tgz -fsSL https://github.com/containernetworking/plugins/releases/download/v1.3.0/cni-plugins-linux-amd64-v1.3.0.tgz
```
pothos, line 94:
```suggestion
curl -o cni.tgz -fsSL https://github.com/containernetworking/plugins/releases/download/v1.3.0/cni-plugins-linux-${ARCH}-v1.3.0.tgz
```
We should also require a second version for this unless you can query the version from some other manifest.

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,99 @@
...
+
+cat > "${SYSEXTNAME}/usr/lib/systemd/system/usr-libexec.mount" <<-'EOF'
+[Unit]
+Description=Kubernetes flex volume plugin directory
+
+[Mount]
+What=overlay
+Where=/usr/libexec
+Type=overlay
+Options=lowerdir=/usr/libexec,workdir=/opt/libexec.work,upperdir=/opt/libexec
```
pothos, line 82: Do you plan to ship anything under `/usr/libexec`? Currently it's not the case, or?

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/bash
+set -euo pipefail
+
+export ARCH="${ARCH-amd64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download the wasmtime release tar ball (e.g., for 4.0.0) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
```
pothos, line 9:
```suggestion
  echo "The script will download the Kubernetes release binaries (e.g., for 1.27.3) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
```

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,99 @@
...
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+rm -f kubectl kubeadm kubelet
+
+# install kubernetes binaries.
+# TODO: use arch variable
```
pothos, line 23:
```suggestion
```

create_kubernetes_sysext.sh
pothos: It also works to download the latest version but I think an explicit parameter is better.

pothos: FYI the github URL for `latest` has it before the `download` part:
https://github.com/containernetworking/plugins/releases/latest/download/cni-plugins-linux-amd64-v1.3.0.tgz

create_kubernetes_sysext.sh
pothos: Ah, understood. I'm a bit concerned about using an overlay because the mounts under `/usr` are manged by systemd-sysext now. If we know what folders should exist and be writable we could ship symlink from `/usr/libexec/kubernetes/` to `/var/kubernetes/` (`/opt` is also not a good idea I think).

pothos: By the way, there is this to join configurations: https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#patches

pothos: With symlink I mean that `/usr/libexec/kubernetes/` would be a symlink to `/var/kubernetes/` and we ensure that `/var/kubernetes/` exists with a helper service. Then the folder is writable.

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+set -euo pipefail
+
+export ARCH="${ARCH-amd64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME CNI_VERSION"
```
pothos, line 8:
```suggestion
  echo "Usage: $0 VERSION SYSEXTNAME [CNI_VERSION]"
```
To indicate it's optional

create_kubernetes_sysext.sh
```diff
@@ -0,0 +1,94 @@
...
+  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
+  echo "CNI version current value is 'latest'"
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+CNI_VERSION="${3-latest}"
+
```
pothos, line 21:
```suggestion
if [ "${ARCH}" = aarch64 ]; then
  ARCH=arm64
fi
```
Sometimes arm64 vs aarch64 is confusing, let's help the user a bit

pothos (approved):

PR #16, opened 2023-08-02
workflow: build and release images

tormath1:
Tested on my fork: https://github.com/tormath1/sysext-bakery/releases/tag/v0.0.1 (https://github.com/tormath1/sysext-bakery/actions/runs/5740907239/job/15559784511)

Tested with systemd-sysupdate:
```yaml
storage:
  files:
    - path: /etc/sysupdate.kubernetes.d/kubernetes.conf
      contents:
        inline: |
          [Transfer]
          Verify=false

          [Source]
          Type=url-file
          Path=https://github.com/tormath1/sysext-bakery/releases/latest/download/
          MatchPattern=kubernetes-@v.raw

          [Target]
          InstancesMax=3
          Type=regular-file
          Path=/opt/extensions/kubernetes
          CurrentSymlink=/etc/extensions/kubernetes.raw
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,47 @@
+name: Build and release Systemd sysext images
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    permissions:
+      # allow the action to create a release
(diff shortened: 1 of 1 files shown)


pothos (approved):
Thanks! Looks good, I would like to also have a systemd-sysupdate config for each component be published as part of the release but that can be done later, too.

PR #21, opened 2023-09-01
Use only systemd arch identifiers for user parameters

pothos:
The difference in architecture identifiers from upstream binary releases and what systemd expects in the extension metadata or the update manifest easily causes confusion.
Recommend the use of the systemd identifiers as they will show up in filenames. Map the identifiers internally and try to do the right thing even if a slighly different identifier was called.

The list of systemd identifiers is here https://www.freedesktop.org/software/systemd/man/systemd.unit.html#ConditionArchitecture=

## Testing done

Built the wasmtime image with the default arch (x86-64) and for arm64.

### DIFF
--- a/create_docker_compose_sysext.sh
+++ b/create_docker_compose_sysext.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -euo pipefail
 
-export ARCH="${ARCH-x86_64}"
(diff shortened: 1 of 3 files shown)


pothos:
We should also add architecture identifiers to the github action, will do it in a new PR

PR #23, opened 2023-09-04
kubernetes: Place cni binaries in /opt/cni/bin

jepio:
# kubernetes: Place cni binaries in /opt/cni/bin

Containerd expects cni binaries in /opt/cni/bin. We need to make sure that directory exists before running the cp, and need to run cp in a way that doesn't result in the source directory being created in target (we only want the executables).

The /opt/libexec thing was an experiment that can be removed as well, as we found another way of dealing with volume plugins.

## How to use

## Testing done

Tested that kubeadm can create a cluster with this.


pothos (approved):

pothos:
We have to make a release afterwards, that's done by pushing a tag for the current date like 20230904.

jepio:
Lets push a tag at the end of the week in case there's any more changes.

PR #24, opened 2023-09-06
Add script to wrap filesystem image in DDI

jepio:
with a dm-verity hash-tree and signed root hash.

This is an example for how to do things for sysexts that will be built into flatcar.


pothos:
I think we should use systemd-repart^^

pothos:
The examples don't cover it because one would also need a third partition with `X-verity-sig` type and `Verity=signature` https://www.freedesktop.org/software/systemd/man/repart.d.html#Examples
The key gets passed in https://www.freedesktop.org/software/systemd/man/systemd-repart.service.html#--private-key=

jepio:
i know you do :)

i'll try that next

pothos:
I'm not against having this script available, too

README.md
```diff
@@ -177,3 +177,14 @@ In case you have an existing Torcx image you can convert it with the `convert_to
 ```
 
 Please make also sure that your don't have a `containerd.service` drop in file under `/etc` that uses Torcx paths.
+
+
+### Verity
+
+To generate sysext protected by dm-verity with a signed root hash pass `FORMAT=verity` before invoking any of the scripts. This requires `systemd-repart` with a version >= v255. This also requires passing a path to a private key and certificate through `KEY` and `CERT`.
```
pothos, line 184: A sentence on how these can be loaded would be good.
```suggestion
To generate sysext protected by dm-verity with a signed root hash pass `FORMAT=verity` before invoking any of the scripts. This requires `systemd-repart` with a version >= v255. This also requires passing a path to a private key and certificate through `KEY` and `CERT`. To load the image, add the certificate file to `/etc/verity.d/`.
```

bake.sh
```diff
@@ -57,7 +72,11 @@ elif [ "${FORMAT}" = "ext4" ] || [ "${FORMAT}" = "ext2" ]; then
   # Note: We didn't chown to root:root, meaning that the file ownership is left as is
   mkfs."${FORMAT}" -E root_owner=0:0 -d "${SYSEXTNAME}" "${SYSEXTNAME}".raw
   resize2fs -M "${SYSEXTNAME}".raw
-else
+elif [ "${FORMAT}" = "squashfs" ]; then
   mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw -all-root
+elif [ "${FORMAT}" = "verity" ]; then
+  systemd-repart --private-key="${KEY}" --certificate="${CERT}" --root="${SYSEXTNAME}" --no-pager --empty=create --size=auto --definitions=repart.d "${SYSEXTNAME}.raw"
```
pothos, line 78: This assumes to be running in the script's folder, we need to add a `cd` or use an absolute path

pothos:
> @pothos: i've added systemd-repart for verity DDI creation to bake.sh

Thanks!

pothos:
From the NEWS entries for 255: `* New --make-ddi=confext, --make-ddi=sysext and --make-ddi=portable options
          have been added to make it easier to generate these types of DDIs,
          without having to provide repart.d definitions for them.`

jepio:
> @jepio @pothos Do you have any news on this? Will it be merged?

I think we hit a roadblock with sysext signature policies: if we enforce signatures for sysexts then all of them need to be signed, but we didn't want to prevent users from loading their own sysexts. And the signing key needs to be built-into the kernel or provided through UEFI mechanisms, which greatly limits mixing Flatcar provided sysext's with user provided ones.

PR #26, opened 2023-09-17
Make script runnable on Alpine, faster Kubernetes builds

href:
Two somewhat unrelated changes that you may or may not care for:

1. Switch to `/usr/bin/env bash`, so that the bakery scripts can be run on Alpine (i.e., inside some Docker containers).
2. Download some Kubernetes dependencies in parallel, to be slightly faster.

## How to use

The easiest way to test both changes, is to create a new Kubernetes sysext build.

## Testing done

I built the Kubernetes sysext on the official Alpine-based Bash image.

### DIFF
--- a/bake.sh
+++ b/bake.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 OS="${OS-flatcar}"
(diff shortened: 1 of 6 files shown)


pothos (approved):
Thank you!

href:
Thank you for merging!

PR #31, opened 2023-10-25
create_docker_sysext.sh: no containerd-shim in v23 and higher

t-lo:
This change fixes a bug with ONLY_DOCKER builds in create_docker_sysext.sh. Docker binary releases version 23 and above do not ship /usr/bin/containerd-shim anymore, leading to the following error when building a "docker only" sysext w/o containerd:

```shell
rm: cannot remove 'docker/usr/bin/containerd-shim': No such file or directory
```

This PR handles versions below major release 23 separately and only deletes containerd-shim for these versions.

Tested with docker releases 20.10.9, 23.0.0, and 24.0.6.


krnowak (approved):
I'm only wondering why we weren't using `rm -f` in the first place.

t-lo:
> I'm only wondering why we weren't using `rm -f` in the first place.

My first approach was to switch to `-f` but then I thought that we deliberately want this to break if docker ships different sets of binaries. Not sure :shrug:

PR #36, opened 2024-01-10
Build all Kubernetes versions available. Fix for https://github.com/flatcar/Flatcar/issues/1293

jhaprins:
# [Title: describe the change in one sentence]

Modify the workflow file to download information about available Kubernetes versions and build all raw images for the last maintenance release for every main release.

## How to use

## Testing done

- I created a fork of this project
- I then did the changes needed for the workflow file to create desired outcome.
- I pushed my changes to github to see if it would result in the images I want to have.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,14 +21,21 @@ jobs:
             curl \
             jq \
             squashfs-tools \
-            xz-utils
+            xz-utils \
+            gawk
+
+          KBS_VERS=$(curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20  https://raw.githubusercontent.com/kubernetes/website/main/data/releases/schedule.yaml | yq -r '.schedules[].previousPatches[0].release' | awk '{print "kubernetes-v"$1}')
+          [[ -z "${KBS_VERS}" ]] && echo "Failed fetching Kubernetes versions" && exit 1
+
+          KBS_VERS_ARRAY=(${KBS_VERS})
 
           images=(
-              "kubernetes-v1.28.2"
               "docker-24.0.6"
               "docker_compose-2.22.0"
               "wasmtime-13.0.0"
           )
+          images+=("${KBS_VERS_ARRAY[@]}")
+
           streams=()
 
           for image in "${images[@]}"; do


.github/workflows/release.yaml
```diff
@@ -21,10 +21,12 @@ jobs:
...
             jq \
             squashfs-tools \
-            xz-utils
+            xz-utils \
+            gawk
+
 
           images=(
-              "kubernetes-v1.28.2"
+              $(curl https://raw.githubusercontent.com/kubernetes/website/main/data/releases/schedule.yaml | yq -r '.schedules[].previousPatches[0].release' | awk '{print "kubernetes-v"$1}')
```
pothos, line 29: Thanks, we need to move the curl call into it's own line for error handling, and also add some common curl options for robustness: `-fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20`

I think we also need to filter out future `targetDate` entries, or?

jhaprins:
I have done the following additional changes:
- I moved the curl call outside of the array
- I added the curl options you requested.
- I check if the resulting string is not empty
- If the string is empty I bail out with an error 1
- I create the array of kubernetes versions
- I add this array to the images array

Jan Hugo Prins

.github/workflows/release.yaml
```diff
@@ -21,14 +21,21 @@ jobs:
...
+
+          KBS_VERS_ARRAY=(${KBS_VERS})
 
           images=(
-              "kubernetes-v1.28.2"
               "docker-24.0.6"
               "docker_compose-2.22.0"
               "wasmtime-13.0.0"
           )
+          images+=(${KBS_VERS_ARRAY[@]})
```
pothos, line 37:
```suggestion
          images+=("${KBS_VERS_ARRAY[@]}")
```

.github/workflows/release.yaml
```diff
@@ -21,14 +21,21 @@ jobs:
             curl \
             jq \
             squashfs-tools \
-            xz-utils
+            xz-utils \
+            gawk
+
+          KBS_VERS=$(curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused --retry-max-time 60 --connect-timeout 20  https://raw.githubusercontent.com/kubernetes/website/main/data/releases/schedule.yaml | yq -r '.schedules[].previousPatches[0].release' | awk '{print "kubernetes-v"$1}')
+          [[ -z "${KBS_VERS}" ]] && exit 1
```
pothos, line 28:
```suggestion
          if [[ -z "${KBS_VERS}" ]]; then; echo "Failed fetching Kubernetes versions"; exit 1; fi
```

.github/workflows/release.yaml
```diff
@@ -21,14 +21,21 @@ jobs:
             curl \
             jq \
             squashfs-tools \
-            xz-utils
+            xz-utils \
+            gawk
```
pothos, line 25: I think we also need to install yq here, is there a package? Otherwise one could also use `grep` ;)

jhaprins: yq is already available. 
Tested and it is working.

pothos:
> I think we also need to filter out future targetDate entries, or?

With that I meant that while the current entries have no future dates that might be the case and then the k8s version can't be downloaded because it's not released yet.

jhaprins:
> With that I meant that while the current entries have no future dates that might be the case and then the k8s version can't be downloaded because it's not released yet.

No, this is not true.
Check for example 1.29.1, this one has a future date but is not included in my selection.
The jq selection rule fetches very specific previousPatches, and therefor it will not fail on next: patches which are not release yet. See: yq -r '.schedules[].previousPatches[0].release'

.github/workflows/release.yaml
jhaprins: Don't really like that if construct. I will make it a little different, but functional the same.
I hope you will agree. 
```suggestion
          [[ -z "${KBS_VERS}" ]] && echo "Failed fetching Kubernetes versions" && exit 1
```

.github/workflows/release.yaml
jhaprins: Agreed. Will modify.

pothos (approved):

PR #40, opened 2024-02-01
release action: Create new release from list, include old release

t-lo:
This change updates the release GitHub Actions workflow to build a new release instead of modifying an existing one. The previous "latest" release will be deleted.

The release will contain all sysexts listed in
"release_build_versions.txt" as well as all sysexts included in the previous latest release. It will generate a Release.md to be included in the release.

The change also updates the workflow to only run when the tag "latest" is pushed.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,7 @@ name: Build and release Systemd sysext images
 on:
   push:
     tags:
-      - '*'
+      - 'latest'
 jobs:
   build:
     runs-on: ubuntu-22.04
@@ -12,8 +12,8 @@ jobs:
     steps:
       # checkout the sources
       - uses: actions/checkout@v3
-      # build the images and generate a manifest
-      - name: build
+      # prepare build host
+      - name: install prerequisites
         run: |
           set -euxo pipefail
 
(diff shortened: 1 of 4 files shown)


t-lo:
(I'm not around for more than a week, feel free to amend and merge this PR as you see fit)

pothos (approved):
Thanks, looks good

.github/workflows/release.yaml
```diff
@@ -2,7 +2,7 @@ name: Build and release Systemd sysext images
 on:
   push:
     tags:
-      - '*'
+      - 'latest'
```
pothos, line 5: Improvement: Also release on a weekly cadence

pothos: Ah, and should we always release `main`? Maybe a new action could run on every push and create a `latest` tag and trigger this action here (I think it won't get invoked by another workflow as this is a mechanism to prevent recursion).

t-lo: Eventually, yes. For this (and the weekly build cadence) to add value we should first make the build auto-detect new versions for all sysexts (like it does for Kubernetes already): I've opened a tracking issue for that: https://github.com/flatcar/sysext-bakery/issues/42

PR #44, opened 2024-02-13
release.yaml: remove spurious checkout action to un-break CI

t-lo:
The "delete tag" action step included a spurious reference to the checkout action which lead to CI breakage:
```
  Error: .github#L1
  a step cannot have both the `uses` and `run` keys
```
This change fixes that issue.


t-lo:
See e.g. https://github.com/flatcar/sysext-bakery/actions/runs/7883802469

krnowak (approved):

PR #46, opened 2024-02-13
release_build: Remove log output and hardcoded date

pothos:
The wget log output is too long and the date needs to be dynamic.

Other topic: I see that an old 202309 release was used to fetch previous assets. Let's delete those.


pothos:
Started a new run to see that it uses the last release: https://github.com/flatcar/sysext-bakery/actions/runs/7890342342

pothos:
Nice how quick it finished :heart_eyes: - also reduces environmental impact :seedling: ;)

PR #48, opened 2024-02-21
bake_flatcar_image.sh: Bake sysexts into Flatcar OS images

t-lo:
This change introduces `bake_flatcar_image.sh`, a standalone script for creating Flatcar OS images with sysexts (from the bakery and other sources) included.

The script will download a generic Flatcar OS image (version can be specified), copy local sysexts into root or OEM partitions, and create respective symlinks to `/etc/extensions`.

By default, the script operates on local sysext files. However, the `--fetch` option may be specified to download  sysext(s) from the latest Bakery release.

Optionally, vendor images can be created. If requested, the script will use the Flatcar SDK corresponding to the OS image version to generate vendor images from the generic image.

**NOTE** that building vendor images requires https://github.com/flatcar/scripts/pull/1682 . To test, apply https://github.com/flatcar/scripts/commit/28070022226970ff3e819516a60721d8648a0f23.diff right after the script checked out the `scripts` repo (can be done in parallel because the script will spend some time downloading SDK container and OS images before continuing).

# How to use

Run the script:
```bash
./bake_flatcar_image.sh --fetch wasmtime:wasmtime-17.0.1-x86-64.raw
```
or
```
wget https://github.com/flatcar/sysext-bakery/releases/download/latest/wasmtime-17.0.1-x86-64.raw
wget https://github.com/flatcar/sysext-bakery/releases/download/latest/wasmtime.conf
./bake_flatcar_image.sh wasmtime:wasmtime-17.0.1-x86-64.raw
```

Then loopback-mount flatcar_production_image.bin (partition 9) to verify.

To build a bootable image, use
```bash
wget https://github.com/flatcar/scripts/commit/28070022226970ff3e819516a60721d8648a0f23.diff
./bake_flatcar_image.sh --vendor qemu wasmtime:wasmtime-17.0.1-x86-64.raw
# AFTER THE SDK WAS CHECKED OUT, in a separate terminal window, run
cd flatcar-os-image
patch -p1 <../28070022226970ff3e819516a60721d8648a0f23.diff
# after the build concluded, run
chmod 755 flatcar_production_qemu_uefi.sh
bunzip2 flatcar_production_qemu_uefi_image.img.bz2
./flatcar_production_qemu_uefi.sh -nographic -snapshot
# [ ... system boots ...]
wasmtime --version
```

# Testing done
- Built wasmtime qemu_uefi OS images for x86-64 and arm64, both `root` and `oem` partitions, booted and tested wasmtime.
- Built ami and azure vendor images but didn't test.

### DIFF
--- a/bake_flatcar_image.sh
+++ b/bake_flatcar_image.sh
@@ -1,0 +1,337 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) 2024 The Flatcar Maintainers.
+# Use of this source code is governed by the Apache 2.0 license.
+#
+# Embed one or more sysexts into a Flatcar generic OS image.
+# Optionally, build a vendor image (requires the Flatcar SDK).
+
+fetch="no"
+vendor="generic"
+release="stable"
+
+# From https://www.freedesktop.org/software/systemd/man/os-release.html#ARCHITECTURE=
+arch="x86-64"
+install_to="root:/opt/extensions/"
+
+set -euo pipefail
+workdir="$(pwd)/flatcar-os-image"
+bakery_base_url="https://github.com/flatcar/sysext-bakery/releases/download/latest"
+
+# ./run_sdk_container ./image_to_vm.sh --help 2>&1 | grep '\--format'
+supported_vendors=( "ami" "ami_vmdk" "azure" "cloudsigma" "cloudstack" "cloudstack_vhd" "digitalocean" "exoscale" "gce" "hyperv" "iso" "openstack" "openstack_mini" "packet" "parallels" "pxe" "qemu" "qemu_uefi" "qemu_uefi_secure" "rackspace" "rackspace_onmetal" "rackspace_vhd" "vagrant" "vagrant_parallels" "vagrant_virtualbox" "vagrant_vmware_fusion" "virtualbox" "vmware" "vmware_insecure" "vmware_ova" "vmware_raw" "xen" )
+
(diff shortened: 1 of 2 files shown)


bake_flatcar_image.sh
```diff
@@ -0,0 +1,337 @@
...
+function create_vendor_image() {
+    local board="$1"
+    local vendor="$2"
+
+    if [ "${vendor}" = "generic" ] ; then
+        return
+    fi
+
+    trap "docker container rm --force flatcar-oem-builder" EXIT
+    COREOS_OFFICIAL=1 ./run_sdk_container -n flatcar-oem-builder \
```
pothos, line 250: Instead of building a new image which requires the SDK and thus takes very long time, we could rather download the released vendor image and use `qemu-nbd` or a temporary `qemu-img convert` step to modify it directly.

t-lo: This doesn't build a new image - it uses the generic image downloaded earlier to create a vendor image. Only the vendor bits are installed, and all of these come from binary packages. It's quite fast.

t-lo: (Also, I'm hesitant to create and to support two different ways of creating vendor images...)

pothos: The step with the binary packages may be fast but the setup of the SDK container not. Anyway, not a blocker.

pothos (approved):

PR #53, opened 2024-03-22
Bake K3S

gcavalcante8808:
# Bake K3S as a systemd sysext image.

This PR aims to bake [K3S](https://k3s.io/) as a systemd sysext and also provide a systemd unit file to manage k3s.

## How to use

## Testing done

### DIFF
--- a/create_k3s_sysext.sh
+++ b/create_k3s_sysext.sh
@@ -1,0 +1,35 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export ARCH="${ARCH-x86-64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download the k3s binary (e.g., for v1.29.2+k3s1) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
+  echo "All files in the sysext image will be owned by root."
+  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+# The github release uses different arch identifiers, we map them here
+# and rely on bake.sh to map them back to what systemd expects
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86-64" ]; then
+  ARCH="x86_64"
+  URL="https://github.com/k3s-io/k3s/releases/download/${VERSION}/k3s"
+elif [ "${ARCH}" = "arm64" ]; then
+  ARCH="aarch64"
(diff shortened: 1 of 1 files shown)


create_k3s_sysext.sh
```diff
@@ -0,0 +1,35 @@
...
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+# The github release uses different arch identifiers, we map them here
+# and rely on bake.sh to map them back to what systemd expects
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86-64" ]; then
+  ARCH="x86_64"
```
pothos, line 23:
```suggestion
```
This can be omitted, since it's not used for the URL

create_k3s_sysext.sh
```diff
@@ -0,0 +1,35 @@
...
+  URL="https://github.com/k3s-io/k3s/releases/download/${VERSION}/k3s"
+elif [ "${ARCH}" = "arm64" ]; then
+  ARCH="aarch64"
```
pothos, line 26:
```suggestion
```

create_k3s_sysext.sh
```diff
@@ -0,0 +1,35 @@
...
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+# The github release uses different arch identifiers, we map them here
+# and rely on bake.sh to map them back to what systemd expects
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86-64" ]; then
+  ARCH="x86_64"
+  URL="https://github.com/k3s-io/k3s/releases/download/${VERSION}/k3s"
+elif [ "${ARCH}" = "arm64" ]; then
```
pothos, line 25:
```suggestion
elif [ "${ARCH}" = "arm64" ] || [ "${ARCH}" = "aarch64" ]; then
```
We try to make it easy to use any of them

create_k3s_sysext.sh
```diff
@@ -0,0 +1,35 @@
...
+  ARCH="aarch64"
+  URL="https://github.com/k3s-io/k3s/releases/download/${VERSION}%2Bk3s1/k3s-arm64"
```
pothos, line 27:
```suggestion
  URL="https://github.com/k3s-io/k3s/releases/download/${VERSION}/k3s-arm64"
```
This suffix already is part of the version, or?

pothos:
Thanks! I made some suggestions but overall it's nice. If you also want to have it built in the releases of this repo, you can add it to `release_build_versions.txt` as `k3s-v1.27.11+k3s1` (or whatever version we should use - ideally we add some logic to discover the latest version but we haven't done that yet for other extensions as well).

pothos (approved):

gcavalcante8808:
@pothos Great! Thanks for the updates on the PR and also thanks to accept!

kastl-ars:
Sorry for the beginner's question, but why does it install to `/usr/local/bin/k3s`? Is it intended to be usable with the install script from k3s.io that will create the service files and such?

Or is this by principle, because `/usr/bin/` should not be modified by sysexts in Flatcar?

pothos:
`/usr/bin/` is the normal place, it's up to the preference of the deployed software/the deployer, like `sbin` vs `bin`. In general one can say that `/usr/local/` would be good to keep things separated but I don't think it makes a big difference (And might even be more confusing when two same binaries land on the system). With `systemd-dissect --mtree` one has a reliable way of seeing sysext contents.

PR #56, opened 2024-04-10
bake_flatcar_image: Fix check for empty sysext list

jepio:
"${sysexts[@]}" expands every argument separately, while we need all array members to be expanded into a single string for the test to work. "${sysexts[*]}" is what we want.


jepio:
Hold on

jepio:
Now it's fixed.

pothos (approved):

PR #60, opened 2024-04-17
bake.sh: Gate mksquashfs xattr exclusion behind version check

pothos:
Old mksquashfs versions don't support the xattr exclusion that was added to prevent harmless warnings getting printed. To support old mksquashfs versions, gate this behind a version check.

## Testing done

I tested `./create_wasmtime_sysext.sh 4.0.0 wasmtime` with a 4.6 version, and for an older version it was also [tested](https://kubernetes.slack.com/archives/C03GQ8B5XNJ/p1713360007210139?thread_ts=1713355624.708779&cid=C03GQ8B5XNJ)

### DIFF
--- a/bake.sh
+++ b/bake.sh
@@ -63,6 +63,13 @@ elif [ "${FORMAT}" = "ext4" ] || [ "${FORMAT}" = "ext2" ]; then
   mkfs."${FORMAT}" -E root_owner=0:0 -d "${SYSEXTNAME}" "${SYSEXTNAME}".raw
   resize2fs -M "${SYSEXTNAME}".raw
 else
-  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw -all-root -noappend -xattrs-exclude '^btrfs.'
+  VER=$({ mksquashfs -version || true ; } | head -n1 | cut -d " " -f 3)
+  VERMAJ=$(echo "${VER}" | cut -d . -f 1)
+  VERMIN=$(echo "${VER}" | cut -d . -f 2)
+  ARG=(-all-root -noappend)
+  if [ "${VERMAJ}" -gt 4 ] && [ "${VERMIN}" -gt 6 ]; then
+    ARG+=('-xattrs-exclude' '^btrfs.')
+  fi
+  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw "${ARG[@]}"
 fi
 echo "Created ${SYSEXTNAME}.raw"


bake.sh
```diff
@@ -63,6 +63,13 @@ elif [ "${FORMAT}" = "ext4" ] || [ "${FORMAT}" = "ext2" ]; then
   mkfs."${FORMAT}" -E root_owner=0:0 -d "${SYSEXTNAME}" "${SYSEXTNAME}".raw
   resize2fs -M "${SYSEXTNAME}".raw
 else
-  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw -all-root -noappend -xattrs-exclude '^btrfs.'
+  VER=$({ mksquashfs -version || true ; } | head -n1 | cut -d " " -f 3)
+  VERMAJ=$(echo "${VER}" | cut -d . -f 1)
+  VERMIN=$(echo "${VER}" | cut -d . -f 2)
+  ARG=()
```
krnowak, line 69: You recently fixed some bash script where empty array was an issue for some old bash version. Maybe let's make sure that we are not dealing with such here too?

```suggestion
  ARG=(-all-root -noappend)
```

bake.sh
```diff
@@ -63,6 +63,13 @@ elif [ "${FORMAT}" = "ext4" ] || [ "${FORMAT}" = "ext2" ]; then
...
+  if [ "${VERMAJ}" -gt 4 ] && [ "${VERMIN}" -gt 6 ]; then
+    ARG+=('-xattrs-exclude' '^btrfs.')
+  fi
+  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw -all-root -noappend "${ARG[@]}"
```
krnowak, line 73:
```suggestion
  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw "${ARG[@]}"
```

bake.sh
```diff
@@ -63,6 +63,13 @@ elif [ "${FORMAT}" = "ext4" ] || [ "${FORMAT}" = "ext2" ]; then
   mkfs."${FORMAT}" -E root_owner=0:0 -d "${SYSEXTNAME}" "${SYSEXTNAME}".raw
   resize2fs -M "${SYSEXTNAME}".raw
 else
-  mksquashfs "${SYSEXTNAME}" "${SYSEXTNAME}".raw -all-root -noappend -xattrs-exclude '^btrfs.'
+  VER=$({ mksquashfs -version || true ; } | head -n1 | cut -d " " -f 3)
+  VERMAJ=$(echo "${VER}" | cut -d . -f 1)
+  VERMIN=$(echo "${VER}" | cut -d . -f 2)
+  ARG=(-all-root -noappend)
+  if [ "${VERMAJ}" -gt 4 ] && [ "${VERMIN}" -gt 6 ]; then
```
ader1990, line 70: The check should be `if [[ "${VERMAJ}" -gt 4 || ( "${VERMAJ}" -eq 4 && "${VERMIN}" -gt 5 ) ]]; then` ?

See:
https://github.com/plougher/squashfs-tools/blob/master/README-4.6.1#L37

krnowak: Yeah, your logic is correct.

PR #65, opened 2024-04-19
Add keepalived

jsalatiel:
Add keepalived sysext
Currently there is no official container for keepalived, so this recipe builds it statically from a github tag.
It requires docker for the building process.

### DIFF
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -21,7 +21,14 @@ if ! which docker &>/dev/null; then
   exit 1
 fi
 
-export ARCH=x86-64
+SUFFIX=
+if  [ "${ARCH}" = "x86-64" ] || [ "${ARCH}" = "x86_64" ]; then
+  ARCH=amd64
+elif [ "${ARCH}" = "arm64" ]; then
+  ARCH="aarch64"
+  SUFFIX="v8"
+fi
+IMG=docker.io/"${ARCH}${SUFFIX}"/alpine:3.19
 mkdir -p "${SYSEXTNAME}"
 cat >"${SYSEXTNAME}"/build.sh <<EOF
 apk --no-cache add \
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -81,7 +81,7 @@ git clone https://github.com/acassen/keepalived.git
     make && \
     make DESTDIR=/install_root install && \
     find /install_root && \
-    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat --format=%u:%g /install_root/build.sh) /install_root -R
+    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat -c %u:%g /install_root/build.sh) /install_root -R
 EOF
 chmod +x "${SYSEXTNAME}"/build.sh
 docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm -it "${IMG}" /bin/sh -c /install_root/build.sh
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -83,7 +83,7 @@ set -ex && \
     rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown $(id -u):$(id -g) /install_root -R
 EOF
 chmod +x "${SYSEXTNAME}"/build.sh
-docker run -v "${SCRIPTFOLDER}"/"${SYSEXTNAME}":/install_root/  --rm -it alpine:3.19 /bin/sh -c /install_root/build.sh
+docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm -it "${IMG}" /bin/sh -c /install_root/build.sh
 mkdir -p  "${SYSEXTNAME}"/usr/lib/systemd/system/
 cat > "${SYSEXTNAME}"/usr/lib/systemd/system/keepalived.service <<-'EOF'
 [Unit]
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -84,7 +84,7 @@ git clone https://github.com/acassen/keepalived.git
     rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat -c %u:%g /install_root/build.sh) /install_root -R
 EOF
 chmod +x "${SYSEXTNAME}"/build.sh
-docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm -it "${IMG}" /bin/sh -c /install_root/build.sh
+docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm "${IMG}" /bin/sh -c /install_root/build.sh
 mkdir -p  "${SYSEXTNAME}"/usr/lib/systemd/system/
 cat > "${SYSEXTNAME}"/usr/lib/systemd/system/keepalived.service <<-'EOF'
 [Unit]
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -16,7 +16,7 @@ fi
 VERSION="$1"
 SYSEXTNAME="$2"
 
-if ! which docker &>/dev/null; then
+if ! command -v docker &>/dev/null; then
   echo Missing docker in path
   exit 1
 fi
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
+export ARCH="${ARCH-x86-64}"
 SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
 
 if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -76,7 +76,6 @@ git clone https://github.com/acassen/keepalived.git
     --enable-nftables \
     --enable-regex \
     --enable-json  --with-init=systemd --enable-vrrp --enable-libnl-dynamic
-set -ex && \
     cd /opt/keepalived && \
     make && \
     make DESTDIR=/install_root install && \
--- a/create_keepalived_sysext.sh
+++ b/create_keepalived_sysext.sh
@@ -80,7 +80,7 @@ set -ex && \
     make && \
     make DESTDIR=/install_root install && \
     find /install_root && \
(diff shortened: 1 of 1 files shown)


create_keepalived_sysext.sh
```diff
@@ -0,0 +1,129 @@
...
+fi
+
+export ARCH=x86-64
+mkdir -p "${SYSEXTNAME}"
+#docker build -t keepalived-build --build-arg  VERSION=$VERSION - <  Dockerfile-keepalived
+docker build -t keepalived-build --build-arg  VERSION=$VERSION - <<-'EOF'
+
+FROM alpine:3.19 as build
+ARG VERSION=not_set_force_fail
+RUN    apk --no-cache add \
```
pothos, line 32: What we could do here is to put all the commands in a temporary "install" script (bash heredoc writing to `${SYSEXTNAME}/install`), and then use `docker run -v "${SYSEXTNAME}":/work /work/install` to run it and prepare the binary in `/work/usr/bin/`.

jsalatiel: I have updated the build process.

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,114 @@
...
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+if ! which docker &>/dev/null; then
+  echo Missing docker in path
+  exit 1
+fi
+
+export ARCH=x86-64
```
pothos, line 24:
```suggestion
SUFFIX=
if  [ "${ARCH}" = "x86-64" ] || [ "${ARCH}" = "x86_64" ]; then
  ARCH=amd64
elif [ "${ARCH}" = "arm64" ]; then
  ARCH="aarch64"
  SUFFIX="v8"
fi
IMG=docker.io/"${ARCH}${SUFFIX}"/alpine:3.19
```

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,114 @@
...
+    --enable-json  --with-init=systemd --enable-vrrp --enable-libnl-dynamic
+set -ex && \
+    cd /opt/keepalived && \
+    make && \
+    make DESTDIR=/install_root install && \
+    find /install_root && \
+    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown $(id -u):$(id -g) /install_root -R
+EOF
+chmod +x "${SYSEXTNAME}"/build.sh
+docker run -v "${SCRIPTFOLDER}"/"${SYSEXTNAME}":/install_root/  --rm -it alpine:3.19 /bin/sh -c /install_root/build.sh
```
pothos, line 79:
```suggestion
docker run -v "${SCRIPTFOLDER}"/"${SYSEXTNAME}":/install_root/  --rm -it "${IMG}" /bin/sh -c /install_root/build.sh
```

jsalatiel: Should I use $(PWD)? What If the script is run from outside the current directory?

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+    --enable-bfd \
+    --enable-nftables \
+    --enable-regex \
+    --enable-json  --with-init=systemd --enable-vrrp --enable-libnl-dynamic
+set -ex && \
+    cd /opt/keepalived && \
+    make && \
+    make DESTDIR=/install_root install && \
+    find /install_root && \
+    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown $(id -u):$(id -g) /install_root -R
```
pothos, line 83: The chown is something we would only do with Docker, not with Podman. Best would be to do the chown to the UID that owns the `build.sh`
```suggestion
    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat --format=%u:%g /install_root/build.sh) /install_root -R
```

create_keepalived_sysext.sh
pothos: The current directory is the one where the script is started from, not where it resides as file, and this is where `mkdir -p "${SYSEXTNAME}"` creates the folder

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+mkdir -p "${SYSEXTNAME}"/usr/lib/systemd/system/keepalived.service.d
+cat > "${SYSEXTNAME}"/usr/lib/systemd/system/keepalived.service.d/10-keepalived.conf <<-'EOF'
+[Service]
+ExecStartPre=/usr/bin/mkdir -p /etc/keepalived/ 
+ExecStartPre=-/bin/bash -c '[ ! -f /etc/keepalived/keepalived.conf ] && touch /etc/keepalived/keepalived.conf'
```
pothos, line 111:
```suggestion
ExecStartPre=/bin/bash -c 'set -e; mkdir -p /etc/keepalived/; if ! [ -e /etc/keepalived/keepalived.conf ]; then touch /etc/keepalived/keepalived.conf; fi' 
```
Using `if` won't propagate the condition as return code and we can drop the `-` in `=-`

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+if  [ "${ARCH}" = "x86-64" ] || [ "${ARCH}" = "x86_64" ]; then
+  ARCH=amd64
+elif [ "${ARCH}" = "arm64" ]; then
+  ARCH="aarch64"
+  SUFFIX="v8"
+fi
+IMG=docker.io/"${ARCH}${SUFFIX}"/alpine:3.19
+mkdir -p "${SYSEXTNAME}"
+cat >"${SYSEXTNAME}"/build.sh <<EOF
+apk --no-cache add \
```
pothos, line 34:
```suggestion
#!/bin/sh
set -euo pipefail
apk --no-cache add \
```

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+        openssl \
+        openssl-dev \
+        openssl-libs-static \
+        pcre2 \
+        pcre2-dev \
+        autoconf \
+        automake zlib-static  alpine-sdk linux-headers libmnl-static git
+cd /opt
+git clone https://github.com/acassen/keepalived.git
+set -ex && \
```
pothos, line 62:
```suggestion
```
And we can also drop all `&& \`  below

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+    --sbindir=/usr/sbin \
+    --sysconfdir=/usr/etc \
+    --datadir=/usr/share \
+    --localstatedir=/var \
+    --mandir=/usr/share/man \
+    --enable-bfd \
+    --enable-nftables \
+    --enable-regex \
+    --enable-json  --with-init=systemd --enable-vrrp --enable-libnl-dynamic
+set -ex && \
```
pothos, line 78:
```suggestion
```
Same here, since we run this as script file we can drop this and all `&& \` below

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+  echo "All files in the sysext image will be owned by root."
+  echo "The build process requires docker"
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+if ! which docker &>/dev/null; then
```
pothos, line 19:
```suggestion
if ! command -v docker &>/dev/null; then
```
Using the shell builtin might be better here

jsalatiel:
The updated version triggers
./create_keepalived_sysext.sh: line 25: ARCH: unbound variable

```
if  [ "${ARCH}" = "x86-64" ] || [ "${ARCH}" = "x86_64" ]; then
```

Does an unbound variable still fails inside a IF?

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
```
pothos, line 4:
```suggestion
export ARCH="${ARCH-x86-64}"
SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
```

pothos: The other scripts have it declared on top because it's something the user can overwrite.

jsalatiel:
stat: unrecognized option: format=%u:%g

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+    --mandir=/usr/share/man \
+    --enable-bfd \
+    --enable-nftables \
+    --enable-regex \
+    --enable-json  --with-init=systemd --enable-vrrp --enable-libnl-dynamic
+    cd /opt/keepalived && \
+    make && \
+    make DESTDIR=/install_root install && \
+    find /install_root && \
+    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat --format=%u:%g /install_root/build.sh) /install_root -R
```
pothos, line 84:
```suggestion
    rm -rf /install_root/usr/share /install_root/usr/etc/keepalived/samples && chown \$(stat -c %u:%g /install_root/build.sh) /install_root -R
```
Oh, busybox doesn't have `--format`

create_keepalived_sysext.sh
pothos: Thanks, works well now. I also wanted to test arm64 but am hitting the Docker hub API limit even though I just downloaded one image so far…

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+EOF
+chmod +x "${SYSEXTNAME}"/build.sh
+docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm -it "${IMG}" /bin/sh -c /install_root/build.sh
```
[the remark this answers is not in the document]
pothos: To support running without a terminal

create_keepalived_sysext.sh
```diff
@@ -0,0 +1,121 @@
...
+if ! command -v docker &>/dev/null; then
+  echo Missing docker in path
+  exit 1
+fi
+
+SUFFIX=
+if  [ "${ARCH}" = "x86-64" ] || [ "${ARCH}" = "x86_64" ]; then
+  ARCH=amd64
+elif [ "${ARCH}" = "arm64" ]; then
+  ARCH="aarch64"
```
pothos, line 29:
```suggestion
elif [ "${ARCH}" = "aarch64" ] || [ "${ARCH}" = "arm64" ]; then
  ARCH="arm64"
```
I got that wrong, now it works

pothos:
Tested the arm64 build and it worked, yet quite slow due to the qemu-user translation.

PR #68, opened 2024-04-26
ADD create_rke2_sysext.sh

johanneskastl:
# add script to create a RKE2 sysext

Add a new script that builds a sysext containing the [Rancher rke2](https://github.com/rancher/rke2/) binary and its systemd units.

## How to use

```bash
./create_rke2_sysext.sh v1.29.2+rke2r1 rke2-v1.29.2+rke2r1
```

This will download rke2 in version `v1.29.2+rke2r1` and create a sysext named `rke2-v1.29.2+rke2r1.raw`.

## Testing done

I tested the script locally and got a working sysext.

### DIFF
--- a/create_rke2_sysext.sh
+++ b/create_rke2_sysext.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export ARCH="${ARCH-x86-64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download the rke2 binary (e.g., for v1.29.2+rke2r1) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
+  echo "All files in the sysext image will be owned by root."
+  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+# The github release uses different arch identifiers, we map them here
+# and rely on bake.sh to map them back to what systemd expects
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86-64" ]; then
+  URL="https://github.com/rancher/rke2/releases/download/${VERSION}/rke2.linux-amd64.tar.gz"
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-amd64.txt"
+elif [ "${ARCH}" = "arm64" ] || [ "${ARCH}" = "aarch64" ]; then
+  URL="https://github.com/rancher/rke2/releases/download/${VERSION}/rke2.linux-arm64.tar.gz"
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-arm64.txt"
+fi
+
+rm -rf "${SYSEXTNAME}"
+mkdir -p "${SYSEXTNAME}/usr/local/"
+
+TMP_DIR="${SYSEXTNAME}/tmp/"
+mkdir -p "${TMP_DIR}"
+curl -o "${TMP_DIR}/rke2.linux-amd64.tar.gz"  -fsSL "${URL}"
+curl -o "${TMP_DIR}/sha256sums"  -fsSL "${SHA256SUMS}"
+pushd "${TMP_DIR}" > /dev/null
+grep rke2.linux-amd64.tar.gz ./sha256sums | sha256sum -c -
+popd  > /dev/null
+
+tar xf "${TMP_DIR}/rke2.linux-amd64.tar.gz" -C "${SYSEXTNAME}/usr/local/"
+rm "${SYSEXTNAME}/usr/local/bin/rke2-uninstall.sh"
+
+# remove TMP_DIR before building the sysext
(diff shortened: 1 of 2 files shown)


create_rke2_sysext.sh
```diff
@@ -0,0 +1,45 @@
...
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-amd64.txt"
+elif [ "${ARCH}" = "arm64" ] || [ "${ARCH}" = "aarch64" ]; then
+  URL="https://github.com/rancher/rke2/releases/download/${VERSION}/rke2.linux-arm64.tar.gz"
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-arm64.txt"
+fi
+
+rm -rf "${SYSEXTNAME}"
+mkdir -p "${SYSEXTNAME}/usr/local/"
+
+TMP_DIR=$(mktemp -d -t rke2-install.XXXXXXXXXX)
```
pothos, line 33:
```suggestion
TMP_DIR="${SYSEXTNAME}/tmp"
```
I think it's better to keep everything in this subfolder so that the user ends with only the subfolder to clean up on a failed build, and the cleanup will also be done on the next run. Using mktemp we would have to install a bash trap.

create_rke2_sysext.sh
```diff
@@ -0,0 +1,45 @@
...
+pushd "${TMP_DIR}" > /dev/null
+grep rke2.linux-amd64.tar.gz ./sha256sums | sha256sum -c -
+popd  > /dev/null
+
+tar xf "${TMP_DIR}/rke2.linux-amd64.tar.gz" -C "${SYSEXTNAME}/usr/local/"
+rm "${SYSEXTNAME}/usr/local/bin/rke2-uninstall.sh"
+
+"${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
+rm -rf "${SYSEXTNAME}"
+rm -rf "${TMP_DIR}"
```
pothos, line 45:
```suggestion
rm -rf "${TMP_DIR}"

"${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
rm -rf "${SYSEXTNAME}"
```
For the above suggestion to work

create_rke2_sysext.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# The github release uses different arch identifiers, we map them here
+# and rely on bake.sh to map them back to what systemd expects
+if [ "${ARCH}" = "amd64" ] || [ "${ARCH}" = "x86-64" ]; then
+  URL="https://github.com/rancher/rke2/releases/download/${VERSION}/rke2.linux-amd64.tar.gz"
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-amd64.txt"
+elif [ "${ARCH}" = "arm64" ] || [ "${ARCH}" = "aarch64" ]; then
+  URL="https://github.com/rancher/rke2/releases/download/${VERSION}/rke2.linux-arm64.tar.gz"
+  SHA256SUMS="https://github.com/rancher/rke2/releases/download/${VERSION}/sha256sum-arm64.txt"
+fi
```
pothos, line 28: FYI: If you want you can only set `ARCH=amd64` and `ARCH=arm64` in the if-else block. Afterwards you can assign `URL=…${ARCH}…`.

create_rke2_sysext.sh
johanneskastl: Good point, fixed in latest commit.

create_rke2_sysext.sh
johanneskastl: fixed in latest commit

create_rke2_sysext.sh
johanneskastl: I just copied the existing k3s script, but if you want me to I can change this?

create_rke2_sysext.sh
```diff
@@ -0,0 +1,49 @@
...
+
+rm -rf "${SYSEXTNAME}"
+mkdir -p "${SYSEXTNAME}/usr/local/"
+
+TMP_DIR="${SYSEXTNAME}/tmp/"
```
pothos, line 33:
```suggestion
TMP_DIR="${SYSEXTNAME}/tmp/"
mkdir -pr "${TMP_DIR}"
```

create_rke2_sysext.sh
pothos: Might be better before it gets copied again and grows further ;)

create_rke2_sysext.sh
johanneskastl: Fixed, although I left out the `-r` as this is not a valid mkdir option...

create_rke2_sysext.sh
johanneskastl: Fixed in latest commit

johanneskastl:
Would it make sense to use `${VARIABLE:?}` to make sure nothing bad happens if for $REASONS a variable is unset?
https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Shell-Parameter-Expansion

johanneskastl:
Do we need the `RELOAD=1` in this one as well?

create_rke2_sysext.sh
pothos: Yes ;)

pothos:
> Would it make sense to use `${VARIABLE:?}` to make sure nothing bad happens if for $REASONS a variable is unset?

The variables are set in the script and if they aren't then the script is wrong. `shellcheck` should tell.

> Do we need the `RELOAD=1` in this one as well?

I see that `rke2-server.service` and `rke2-agent.service` are included, so yes.
And we also need the drop-in (see https://github.com/flatcar/sysext-bakery/pull/69#discussion_r1582889217) for them to be started on load.

johanneskastl:
reload see b4e62e6

johanneskastl:
Rebased on latest state in `main`, added README section for rke2 (and headings to better structure the sections with wasmcloud, k3s and rke)

pothos (approved):

PR #69, opened 2024-04-27
create_k3s_sysext.sh: create systemd unit files

johanneskastl:
#  create systemd unit files

create `/usr/local/lib/systemd/system/k3s.service` and `/usr/local/lib/systemd/system/k3s-agent.service`

unit files were taken from the k3s install script at [get.k3s.io](https://get.k3s.io/).

### DIFF
--- a/create_k3s_sysext.sh
+++ b/create_k3s_sysext.sh
@@ -95,5 +95,5 @@ ExecStartPre=-/sbin/modprobe overlay
 ExecStart=/usr/local/bin/k3s agent
 EOF
 
-"${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
+RELOAD=1 "${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
 rm -rf "${SYSEXTNAME}"
--- a/create_k3s_sysext.sh
+++ b/create_k3s_sysext.sh
@@ -29,5 +29,71 @@ rm -rf "${SYSEXTNAME}"
 mkdir -p "${SYSEXTNAME}"/usr/local/bin
 curl -o "${SYSEXTNAME}/usr/local/bin/k3s" -fsSL "${URL}"
 chmod +x "${SYSEXTNAME}"/usr/local/bin/k3s
+
+mkdir -p "${SYSEXTNAME}"/usr/local/lib/systemd/system/
+cat > "${SYSEXTNAME}"/usr/local/lib/systemd/system/k3s.service << EOF
+[Unit]
+Description=Lightweight Kubernetes
+Documentation=https://k3s.io
+Wants=network-online.target
+After=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=notify
+EnvironmentFile=-/etc/default/%N
+EnvironmentFile=-/etc/sysconfig/%N
+EnvironmentFile=-/usr/local/lib/systemd/system/k3s.service.env
+KillMode=process
+Delegate=yes
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
+LimitNOFILE=1048576
+LimitNPROC=infinity
+LimitCORE=infinity
+TasksMax=infinity
+TimeoutStartSec=0
+Restart=always
+RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service 2>/dev/null'
+ExecStartPre=-/sbin/modprobe br_netfilter
+ExecStartPre=-/sbin/modprobe overlay
+ExecStart=/usr/local/bin/k3s server
+EOF
+
+cat > "${SYSEXTNAME}"/usr/local/lib/systemd/system/k3s-agent.service << EOF
+[Unit]
+Description=Lightweight Kubernetes
+Documentation=https://k3s.io
+Wants=network-online.target
+After=network-online.target
+
+[Install]
+WantedBy=multi-user.target
+
+[Service]
+Type=notify
+EnvironmentFile=-/etc/default/%N
+EnvironmentFile=-/etc/sysconfig/%N
+EnvironmentFile=-/usr/local/lib/systemd/system/k3s-agent.service.env
+KillMode=process
+Delegate=yes
+# Having non-zero Limit*s causes performance problems due to accounting overhead
+# in the kernel. We recommend using cgroups to do container-local accounting.
+LimitNOFILE=1048576
+LimitNPROC=infinity
+LimitCORE=infinity
+TasksMax=infinity
+TimeoutStartSec=0
+Restart=always
+RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service 2>/dev/null'
+ExecStartPre=-/sbin/modprobe br_netfilter
+ExecStartPre=-/sbin/modprobe overlay
+ExecStart=/usr/local/bin/k3s agent
+EOF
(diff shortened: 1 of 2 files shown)


create_k3s_sysext.sh
```diff
@@ -29,5 +29,71 @@ rm -rf "${SYSEXTNAME}"
...
+mkdir -p "${SYSEXTNAME}"/usr/local/lib/systemd/system/
+cat > "${SYSEXTNAME}"/usr/local/lib/systemd/system/k3s.service << EOF
+[Unit]
+Description=Lightweight Kubernetes
+Documentation=https://k3s.io
+Wants=network-online.target
+After=network-online.target
+
+[Install]
+WantedBy=multi-user.target
```
pothos, line 42:
```suggestion
UpheldBy=multi-user.target
```
This works with systemd 255 when you pre-create a symlink with `.upholds/`. Otherwise you can keep `Wantedby` for systemd 252 and create a multi-user-target drop in that sets `Upholds=k3s.service`.

Same for the agent service.

pothos: For reference, see `create_kubernetes_sysext.sh`

create_k3s_sysext.sh
```diff
@@ -29,5 +29,71 @@ rm -rf "${SYSEXTNAME}"
...
+TimeoutStartSec=0
+Restart=always
+RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service 2>/dev/null'
+ExecStartPre=-/sbin/modprobe br_netfilter
+ExecStartPre=-/sbin/modprobe overlay
+ExecStart=/usr/local/bin/k3s agent
+EOF
+
 "${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
```
pothos, line 98:
```suggestion
RELOAD=1 "${SCRIPTFOLDER}"/bake.sh "${SYSEXTNAME}"
```
We need to trigger a `systemctl daemon-reload`.

create_k3s_sysext.sh
jepio: don't we ship "systemctl restart multi-user.target` in `ensure-sysext.service` so that users don't have to complicate their sysexts with "Upholds"/"UpheldBy"?

johanneskastl: I'll wait for a final answer here, before adapting.

pothos: > don't we ship `systemctl restart multi-user.target` in `ensure-sysext.service` so that users don't have to complicate their sysexts with "Upholds"/"UpheldBy"?

Yes, but that was a workaround until we have a proper way to do this upstream with `.upholds` symlinks that `systemctl enable` will create for `UpheldBy`. The target restart is sometime we want to get rid of because it might create a race condition. Later on when we would improve the overlay setup to be done earlier the `.wants` might work again if the extension is loaded at boot time but not for extensions added/reloaded later. Also, the extensions are build for `OS=_any` and other distros would not ship the Flatcar workaround.

johanneskastl: Thanks for the clarification, fixed in 289f38e

create_k3s_sysext.sh
johanneskastl: fixed in c91bfe0

create_k3s_sysext.sh
```diff
@@ -29,5 +29,71 @@ rm -rf "${SYSEXTNAME}"
...
+TasksMax=infinity
+TimeoutStartSec=0
+Restart=always
+RestartSec=5s
+ExecStartPre=/bin/sh -xc '! /usr/bin/systemctl is-enabled --quiet nm-cloud-setup.service 2>/dev/null'
+ExecStartPre=-/sbin/modprobe br_netfilter
+ExecStartPre=-/sbin/modprobe overlay
+ExecStart=/usr/local/bin/k3s agent
+EOF
+
```
pothos, line 97:
```suggestion

mkdir -p "${SYSEXTNAME}/usr/local/lib/systemd/system/multi-user.target.d"
{ echo "[Unit]"; echo "Upholds=k3s.service"; } > "${SYSEXTNAME}/usr/local/lib/systemd/system/multi-user.target.d/10-k3s-service.conf"
{ echo "[Unit]"; echo "Upholds=k3s-agent.service"; } > "${SYSEXTNAME}/usr/local/lib/systemd/system/multi-user.target.d/10-k3s-agent-service.conf"

```
We also need to pre-enable them (Here with a drop-in instead of the symlink to be compatible with systemd < 255).

pothos: (Unless one is for the worker, one for the controller, but then one could also mask the unit per Ignition to keep the config there smaller)

johanneskastl: Hmmm, my line of thinking was for the sysext to just put everything in place and the user then having to manually enable/start whatever she likes. It might also require some configuration before the services can be started...

pothos: The configuration is normally brought in via Ignition and already present before the service starts. In other scenarios if the service errors out with missing configuration it's also ok to leave it restarting until the configuration is in place.
At least this is what we did with the Kubernetes/kubeadm systemd-sysext image.

pothos: If we are unsure, we can leave the services disabled and document how to enable them.

johanneskastl: The issue is that k3s starts up fine even without configuration, but it might

* create a new cluster while you wanted it to be an agent or
* it might create a new cluster with settings that you did not want

So I would vote for leaving them disabled, but somehow tell the user how to enable them.

pothos: In the README we can put this Butane config here:
```
variant: flatcar
version: 1.0.0
storage:
  links:
    - path: /etc/systemd/system/multi-user.target.wants/k3s.service
      target: /usr/local/lib/systemd/k3s.service
      overwrite: true
    - path: /etc/systemd/system/multi-user.target.wants/k3s-agent.service
      target: /usr/local/lib/systemd/k3s-agent.service
      overwrite: true
```
We can also switch back to `WantedBy` to make `systemctl enable` work on systemd 252 which is still the version used in Flatcar Stable.

johanneskastl: I dropped the commit that changed the `WantedBy`. So this should be fine now.

Which README do you mean? The general one or is there a README for the k3s sysext?

pothos: Thanks, in the general README, before/after the wasmcloud section

johanneskastl: Done, see commit 0f46e4c

pothos (approved):

kastl-ars:
I just had some time to actually test the k3s sysext, which fails:

```
$ systemd-sysext refresh
Failed to read metadata for image k3s-v1.29.2+k3s1-x86-64: No medium found
$ ls -lh /etc/extensions/
total 56M
lrwxrwxrwx. 1 root root   9 May  7 13:46 containerd-flatcar.raw -> /dev/null
lrwxrwxrwx. 1 root root   9 May  7 13:46 docker-flatcar.raw -> /dev/null
-rw-r--r--. 1 root root 56M May  7 13:46 k3s-v1.29.2+k3s1-x86-64.raw
lrwxrwxrwx. 1 root root  33 May  7 13:46 oem-qemu.raw -> /oem/sysext/oem-qemu-3815.2.2.raw
$
```

Should I open up an issue?

pothos:
The file name must be `/etc/extensions/k3s.raw` which can be a symlink to a folder that has versioned files as done in the README and managed by systemd-sysupdate.

kastl-ars:
Ah, thanks @pothos, that works. I'll prepare a PR for the README, to make the k3s and rke2 sections more self-contained, with all the bits and pieces needed.

PR #70, opened 2024-05-04
feat: add a tailscale bake

guilhem:
Add tailscale sysext

### DIFF
--- a/create_tailscale_sysext.sh
+++ b/create_tailscale_sysext.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export ARCH="${ARCH-x86-64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download tailscale binaries (e.g., for 1.64.0) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "A temporary directory named SYSEXTNAME in the current folder will be created and deleted again."
+  echo "All files in the sysext image will be owned by root."
+  echo "To use arm64 pass 'ARCH=arm64' as environment variable (current value is '${ARCH}')."
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+if [ "${ARCH}" = "x86-64" ]; then
+  ARCH="amd64"
+elif [ "${ARCH}" = "aarch64" ]; then
+  ARCH="arm64"
+fi
+
+TARBALL="tailscale_${VERSION}_${ARCH}.tgz"
+URL="https://pkgs.tailscale.com/stable/${TARBALL}"
+
+rm -rf "${SYSEXTNAME}"
+TMP_DIR="${SYSEXTNAME}/tmp"
+mkdir -p "${TMP_DIR}"
+
+curl -o "${TMP_DIR}/${TARBALL}" -fsSL "${URL}"
+
+tar xf "${TMP_DIR}/${TARBALL}" -C "${TMP_DIR}" --strip-components=1
+
+mkdir -p "${SYSEXTNAME}"/usr/local/{bin,sbin,lib/{systemd/system,extension-release.d}}
+
+mv "${TMP_DIR}/tailscale" "${SYSEXTNAME}/usr/local/bin/tailscale"
+mv "${TMP_DIR}/tailscaled" "${SYSEXTNAME}/usr/local/sbin/tailscaled"
+mv "${TMP_DIR}/systemd/tailscaled.service" "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+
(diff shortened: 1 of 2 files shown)


create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export ARCH="${ARCH-x86-64}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 2 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 VERSION SYSEXTNAME"
+  echo "The script will download tailscale binaries and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
```
pothos, line 9:
```suggestion
  echo "The script will download tailscale binaries (e.g., for 1.64.0) and create a sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
```
We include an example version number to make the format clear (`v` prefix or not) and make testing easy without searching for a version to use.

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+  "${SCRIPTFOLDER}"/bake.sh --help
+  exit 1
+fi
+
+VERSION="$1"
+SYSEXTNAME="$2"
+
+if [ "${ARCH}" = "x86-64" ]; then
+  ARCH="amd64"
+fi
```
pothos, line 22:
```suggestion
elif [ "${ARCH}" = "aarch64" ]; then
  ARCH="arm64"
fi
```
We try to make it easy to use any valid arch specifier.

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+
+tarball="$(curl -fsSL "https://pkgs.tailscale.com/stable/?v=${VERSION}&mode=json" | jq -r ".Tarballs.${ARCH}")"
+
+URL="https://pkgs.tailscale.com/stable/${tarball}"
```
pothos, line 26:
```suggestion
TARBALL="tailscale_${VERSION}_${ARCH}.tgz"
URL="https://pkgs.tailscale.com/stable/${TARBALL}"
```

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+
+TMP_DIR="$(mktemp -d)"
+
+#trap 'rm -rf "${TMP_DIR}"' EXIT
```
pothos, line 30:
```suggestion
TPMDIR="${SYSEXTNAME}/tmp"
mkdir -p "${TMPDIR}"
```
And when we remove this before the `bake.sh` call we don't need a trap.

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+
+curl -o "${TMP_DIR}/tailscale-amd64.tar.gz" -fsSL "${URL}"
```
pothos, line 32:
```suggestion
curl -o "${TMP_DIR}/${TARBALL}" -fsSL "${URL}"
```

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+
+mkdir -p "${TMP_DIR}/extracted"
```
pothos, line 34: I think we can directly use `TMP_DIR` without the `extracted` subfolder

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+tar xf "${TMP_DIR}/tailscale-amd64.tar.gz" -C "${TMP_DIR}/extracted" --strip-components=1
+
+rm -rf "${SYSEXTNAME}"
```
pothos, line 37: This needs to be moved to the start

create_tailscale_sysext.sh
```diff
@@ -0,0 +1,48 @@
...
+
+mkdir -p "${SYSEXTNAME}"/usr/local/{bin,sbin,lib/{systemd/system,extension-release.d}}
+
+mv "${TMP_DIR}/extracted/tailscale" "${SYSEXTNAME}/usr/local/bin/tailscale"
+mv "${TMP_DIR}/extracted/tailscaled" "${SYSEXTNAME}/usr/local/sbin/tailscaled"
+mv "${TMP_DIR}/extracted/systemd/tailscaled.service" "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+
+sed -i 's/--port.*//g' "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+
```
pothos, line 46: Do you want to make the service start on load or only when the user explicitly has configured it? Normally it's ok to make it start on load when it errors out on missing configuration.

guilhem:
I think with that many suggestions, it's ok to commit the way you want it directly to my branch :)
I'm allowing this specifically for this ;)
image

This PR come mainly from this tailscale blog post https://tailscale.com/blog/steam-deck#systemd-sysext :)

create_tailscale_sysext.sh
guilhem: I don't really know what is the best way :/
I don't have enough experience with sysext in production to know what will be better.

pothos: In the README we have an example on how to enable the service if the sysext image doesn't pre-enable it. We can leave it disabled if unsure.

guilhem: This PR will be a first shoot.
Let's see what will be the usage.

pothos:
Thanks, added a readme section as well.

PR #72, opened 2024-05-08
Add example configuration for k3s and rke2 to download the sysext files and place them properly

johanneskastl:


pothos:
Thanks, maybe a note that sysupdate is not use in the snippet, i.e., no updates - which I think makes even sense because updating the kubelet doesn't work without manual hand-holding across major versions.

johanneskastl:
> Thanks, maybe add a note that sysupdate is not used in the snippet, i.e., no updates - which I think makes even sense because updating the kubelet doesn't work without manual hand-holding across major versions.

Done.

PR #73, opened 2024-05-08
create_k3s_sysext.sh: create symlinks 

johanneskastl:
Create symlinks for kubectl, ctr and crictl, like the k3s install script does

### DIFF
--- a/create_k3s_sysext.sh
+++ b/create_k3s_sysext.sh
@@ -29,6 +29,11 @@ rm -rf "${SYSEXTNAME}"
 mkdir -p "${SYSEXTNAME}"/usr/local/bin
 curl -o "${SYSEXTNAME}/usr/local/bin/k3s" -fsSL "${URL}"
 chmod +x "${SYSEXTNAME}"/usr/local/bin/k3s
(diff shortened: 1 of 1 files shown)


create_k3s_sysext.sh
```diff
@@ -29,6 +29,11 @@ rm -rf "${SYSEXTNAME}"
 mkdir -p "${SYSEXTNAME}"/usr/local/bin
 curl -o "${SYSEXTNAME}/usr/local/bin/k3s" -fsSL "${URL}"
 chmod +x "${SYSEXTNAME}"/usr/local/bin/k3s
+pushd "${SYSEXTNAME}"/usr/local/bin/
+ln -s ./k3s kubectl
+ln -s ./k3s ctr
+ln -s ./k3s crictl
+popd
```
pothos, line 36: Is this also needed for the rke2 extension?

johanneskastl: I am currently checking this...

johanneskastl: Apparently not:
https://get.rke2.io/ contains no hint. And the rke2 shell script unpacks a tarball, which apparently does not contain the links.

PR #74, opened 2024-05-08
flix and flatwrap: Add tools to convert a chroot to a sysext

pothos:
Existing binaries can be bundled in a sysext image if they use their own library path or do what Flatpak does and run in a chroot. This introduces the Flix tool which creates sysext images from a given chroot and a set of binaries in it, and then patches the ELF rpath section and to refer to a custom library path.
The other tool introduced here is the Flatwrap tool that creates sysext images from a given chroot and a set of binaries in it, and sets up wrapper entry points with bwrap/unshare into the chroot. This also introduces a helper oci-rootfs.sh to create a useable chroot from a container image.

## How to use/Testing done

See README

### DIFF
--- a/flatwrap.sh
+++ b/flatwrap.sh
@@ -1,0 +1,137 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+export ARCH="${ARCH-amd64}"
+KEEP="${KEEP-}"
+ETCMAP="${ETCMAP-host}"
+VARMAP="${VARMAP-host}"
+HOMEMAP="${HOMEMAP-host}"
+CHROOT="${CHROOT-/usr /lib /lib64 /bin /sbin}"
+HOST="${HOST-/dev /proc /sys /run /tmp /var/tmp}"
+SCRIPTFOLDER="$(dirname "$(readlink -f "$0")")"
+
+if [ $# -lt 3 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
+  echo "Usage: $0 FOLDER SYSEXTNAME PATHS..."
+  echo "The script will set up entry points for the specified binary or systemd unit paths (e.g., /usr/bin/nano, /usr/systemd/system/my.service) from FOLDER into a chroot under /usr/local/, and create a systemd-sysext squashfs image with the name SYSEXTNAME.raw in the current folder."
+  echo "Paths under /usr are recommended but paths under /etc or /bin can also be specified as 'CHROOT:TARGET', e.g., '/etc/systemd/system/my.service:/usr/systemd/system/my.service' or '/bin/mybin:/usr/bin/mybin' supported."
+  echo "Since only the specified paths are available in the host, any files accessed by systemd for service units must also be specified."
+  echo "The binary itself will be able to access all files of the chroot as specifed in the CHROOT environment variable (current value is '${CHROOT}')."
+  echo "It will also be able to access all files of the host as specified in the HOST environment variable and to /etc, /var, /home if not disabled below (current value is '${HOST}')."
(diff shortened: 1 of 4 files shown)


flatwrap.sh
```diff
@@ -0,0 +1,109 @@
...
+    ln -fs --no-target-directory "/usr/local/${SYSEXTNAME}/${ENTRY}" "${SYSEXTNAME}/${DIR}/${NAME}"
+  else
+      tee "${SYSEXTNAME}/${ENTRY}" > /dev/null <<EOF
+#!/bin/sh
+if [ "\$(id -u)" = 0 ]; then
+  exec unshare -m "\${FLATWRAP_ROOT-}/usr/local/${SYSEXTNAME}/unshare-helper" "${ENTRY}" "\$@"
+elif command -v bwrap >/dev/null && [ "\${NOBWRAP}" = "" ]; then
+  exec bwrap ${BWCARGS[@]} ${BWHARGS[@]} "${ENTRY}" "\$@"
+else
+  exec unshare -m -U -c --root "\${FLATWRAP_ROOT-}/usr/local/${SYSEXTNAME}" "${ENTRY}" "\$@"
```
pothos, line 99: I noticed that unpriv ovelay mounts would work to emulate the bind mount. Read-only case would be `mount -t overlayfs -t overlay overlay -o lowerdir=/source:/somethingempty /target`. The read-only case would also be a fallback when the write case fails because that needs to have a writable working dir on the same filesystem as the source `mount -t overlayfs -t overlay overlay -o lowerdir=/somethingempty,upperdir=/source,workdir=/path-on-source-fs /target` (The work dir must not contain the upperdir and vice versa, and for the work dir a `work` sub directory will get created).

Since in the unpriv. case the command wouldn't have write permissions anyway for `/var` or `/etc` it should be ok to use the read-only fallback mount. For `/home` it's a bit special because it normally contains the users's home which is writable but the rest is not, and one could special case mounting `$HOME` but even then one needs a writable work dir that is not in `$HOME`. `/var/tmp` might be ok if it's not a tmpfs. What could work is mapping the subfolders of `$HOME` at the cost of having to create a work dir for each like `$HOME/.cache/SUBFOLDER-work`. I think I would go with a read-only mount for the unpriv. case, maybe add a fresh tmpfs for `/tmp` and just require `bwrap` for anything more special.

pothos: Alternative idea: Don't do the chroot step but prepare the original `/` as needed, i.e., bind mount the read-only sysext contents over `/usr` and other things from `CHROOT` while we either ignore HOST or kick out unwanted dirs by mounting an empty dir over it (needs two empty different dirs at hand in the sysext).

The advantage is that the host dirs stay in place as wanted but a challenge is that the mount command runs "live" and the final unshare command, too, so that one might need to keep the a reference to the host binaries around: mount host `/usr` to `/host/usr` and then run `/host/usr/lib/ld-linux-x86-64.so.2 --library-path /host/… mount -t …` and same for `unshare --map-user=…`.

pothos: Turns out I was doing it wrong and should use `mount --rbind` instead of `mount --bind`. Then the user namespace can do bind mounts. Will update the code.

pothos: Done: When bwrap is not available things should work the same through the unshare commands. I don't even know if there is a benefit of using bwrap because this also seems to lose the mapping of other users.

pothos:
Relaxed the constraint that one has to have a merged `/usr` in the chroot. Now one can specify thing in a separate `/bin` (or `/etc/systemd/system/`) by mapping it to `/usr/bin` (or `/usr/lib/systemd/system/`).

PR #78, opened 2024-06-06
Add a sysext for vim-huge

jsalatiel:
There are several cases when the user needs to edit k8s yamls inplace directly on the master nodes and having syntax highlighting and advanced formatting of vim available is a nice feature since the default vim of flatcar have those disabled.

### DIFF
--- a/create_vim-huge_sysext.sh
+++ b/create_vim-huge_sysext.sh
@@ -45,5 +45,32 @@ EOF
 chmod +x "${SYSEXTNAME}"/build.sh
 docker run -v "${PWD}/${SYSEXTNAME}":/install_root/  --rm "${IMG}" /bin/sh -c /install_root/build.sh
 rm -f "${SYSEXTNAME}"/build.sh
+cat > "${SYSEXTNAME}"/usr/share/vim/vimrc  <<-"EOF"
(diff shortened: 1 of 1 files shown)


chewi:
We have just discussed whether we might want to leverage existing Gentoo packages for this instead. We could potentially do this in such a way that avoids dependence on system libraries (besides glibc) either by still statically linking or by placing the dynamic libraries within the sysext. I could look into this soon, but I don't mind accepting this in the interim.

pothos:
FYI, the build process of the extensions in this repo is supposed to be fast, avoiding compilation if we can (that's the reason one build recipe was not included in the official repo releases so far).
One can bundle existing distro binaries (e.g., from a Docker image) with one of these helper scripts:
https://github.com/flatcar/sysext-bakery?tab=readme-ov-file#flix-and-flatwrap - for choosing one way and making it work one needs to know what files vim relies on at runtime.

That said, I'm not reviewing things here now and someone else has to decide what to do with this PR.

xinity:
IMHO sysext should not be considered as the next package manager, so perhaps vim as an editor that exists in every linux distribution might not fit here.

just sharing my point of view, no harm :)

chewi:
@pothos is away for an extended period. Unfortunately, I still don't know any more about this than I did in August. @tormath1, does this look okay to you?

Issue #42, opened 2024-02-12
[RFE] Auto-detect latest release versions in release builds


t-lo:
## Current situation

While new Kubernetes patch-level releases of all supported minor versions are auto-detected and the latest one is built, a similar feature is not implemented for docker, docker-compose, and wasmtime.

## Impact

New versions need to be specified manually in order to be included in a release build.

## Ideal future situation

New versions are automatically detected by release builds.
Ideally, the implementation is modular and makes it easy to add new components / sysexts to the release auto-build.

## Follow-up tasks
- build new releases on a schedule (e.g. daily, weekly, etc) to keep "latest" release up to date with upstream releases.
- ~automatically tag "latest" in the build instead of relying on manual tagging.~ Done. Release CI action will re-tag.

t-lo:
Implemented in #115 , #117 . Closing.

PR #83, opened 2024-08-02
release-build: build cri-o for each Kubernetes version

tormath1:
CRI-O is coupled to Kubernetes for the releases, while the major and the minor follows Kubernetes streams, the patch version is independant, so we need to fetch the latest patch release for each minor Kubernetes version.

sysupdate configuration is generated similar to Kubernetes. 

---

Locally tested, will trigger a release right after to generate cri-o missing images.

Related to: https://github.com/flatcar/sysext-bakery/issues/42

### DIFF
--- a/release_build.sh
+++ b/release_build.sh
@@ -15,8 +15,7 @@ echo "================================================="
 KBS_VERS=$(curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused \
                 --retry-max-time 60 --connect-timeout 20  \
                 https://raw.githubusercontent.com/kubernetes/website/main/data/releases/schedule.yaml \
-		| yq -r '.schedules[] | .previousPatches[0] // (.release = .release + ".0") | .release' \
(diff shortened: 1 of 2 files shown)


release_build.sh
```diff
@@ -25,6 +24,16 @@ fi
 KBS_VERS_ARRAY=(${KBS_VERS})
 printf "%s\n" "${KBS_VERS_ARRAY[@]}"
 
+echo "Fetching list of latest CRI-O patch releases"
+echo "================================================="
+git clone https://github.com/cri-o/cri-o
```
krnowak, line 29: You could get a sorted list of tags without cloning using `git ls-remote --tags --sort=-v:refname https://github.com/cri-o/cri-o`. You probably could store the output in the file and then process it further to get the latest point release for each kubernetes major version.

But your solution works too, so just saying.

krnowak (approved):

PR #87, opened 2024-09-03
release: do not fail if there is no release artifact

tormath1:
if there is no release artifact, grep command will return an exit-code different from 0 so the script is failing. (https://github.com/flatcar/sysext-bakery/actions/runs/10685052349/job/29616925462)

---

For an unknown reason yet the last release has failed - but it's impossible to rerun has the release script expect to have existing artifacts already released.


chewi (approved):
jq should probably do the filtering but okay.

jepio:
This might actually be rate-limiting (i think i hit that as well), so we would rather want to sleep and retry.

PR #92, opened 2024-10-01
Rework sysext release

jepio:
Fix issues we have had with sysext releases. Previously upload failure would cause us to lose release assets, because the release gets deleted before upload.


jepio:
@t-lo we need this before we push the next "latest" tag. Mathieu faced issues with previous releases.

t-lo (approved):
LGTM, thank you for the quick turn-around, Jeremi!

PR #96, opened 2024-10-10
bake: fix mksquashfs version check

ader1990:
The mksquashfs version that supports the xattrs-exclude flag is 4.6.1, but the logic excluded the version because of a strict (greater than) comparison.

See: https://github.com/flatcar/sysext-bakery/pull/60/commits/d1419fc1254f9723a3614811bdd29f9d6a3bba9d

## Testing done

### DIFF
--- a/bake.sh
+++ b/bake.sh
@@ -38,6 +38,10 @@ elif [ "${ARCH}" = "aarch64" ]; then
   ARCH="arm64"
 fi
 
+function version_ge() {
+    test "$(printf '%s\n' "$@" | sort -rV | head -n 1)" == "$1";
+}
+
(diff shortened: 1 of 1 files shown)


bake.sh
```diff
@@ -38,6 +38,10 @@ elif [ "${ARCH}" = "aarch64" ]; then
   ARCH="arm64"
 fi
 
+function version_ge() {
+    test "$(echo "$@" | tr " " "\n" | sort -rV | head -n 1)" == "$1";
```
krnowak, line 42: How about this:

```suggestion
    test "$(printf '%s\n' "$@" | sort -rV | head -n 1)" == "$1";
```

ader1990: solved, thanks!

Issue #105, opened 2024-12-26
tailscaled.service cannot start due to incorrect path


jmacdonagh:
## Description

The tailscale sysext uses the vendor provided tailscaled.service which has `ExecStart=/usr/sbin/tailscaled ...`, but the binaries are "installed" to `/usr/local/sbin`. When attempting to start the service it fails to start with the error `Job for tailscaled.service failed because of unavailable resources or another system error.`

## Impact

Tailscale sysext is non-functional.

## Environment and steps to reproduce

1. **Set-up**: Provision flatcar with the following butane config
```yaml
variant: flatcar
version: 1.0.0
storage:
  files:
    - path: /opt/extensions/tailscale/tailscale-1.76.6-x86-64.raw
      contents:
        source: https://github.com/flatcar/sysext-bakery/releases/download/latest/tailscale-1.76.6-x86-64.raw
    - path: /etc/sysupdate.d/noop.conf
      contents:
        source: https://github.com/flatcar/sysext-bakery/releases/download/latest/noop.conf
    - path: /etc/sysupdate.tailscale.d/tailscale.conf
      contents:
        source: https://github.com/flatcar/sysext-bakery/releases/download/latest/tailscale.conf
  links:
    - target: /opt/extensions/tailscale/tailscale-1.76.6-x86-64.raw
      path: /etc/extensions/tailscale.raw
      hard: false
systemd:
  units:
    - name: systemd-sysupdate.timer
      enabled: true
    - name: systemd-sysupdate.service
      dropins:
        - name: tailscale.conf
          contents: |
            [Service]
            ExecStartPre=/usr/lib/systemd/systemd-sysupdate -C tailscale update
        - name: sysext.conf
          contents: |
            [Service]
            ExecStartPost=systemctl restart systemd-sysext
```
3. **Task**: Attempt to start service: `sudo systemctl start tailscaled.service`
4. **Action(s)**: N/A
5. **Error**:
```
$ sudo systemctl start tailscaled.service
Job for tailscaled.service failed because of unavailable resources or another system error.
See "systemctl status tailscaled.service" and "journalctl -xeu tailscaled.service" for details.
```

## Expected behavior

Tailscale service should start

## Additional information

```
$ grep ExecStart /usr/local/lib/systemd/system/tailscaled.service
ExecStart=/usr/sbin/tailscaled --state=/var/lib/tailscale/tailscaled.state --socket=/run/tailscale/tailscaled.sock
[...]

PR #106, opened 2025-01-04
Fix tailscale

jmacdonagh:
# Fix tailscale sysext so the service can start by default

The tailscale sysext had a number of issues as described in #105. Specifically:
- The binaries were placed in a different directory than the vendor supplied systemd service expected
- There was no `EnvironmentFile`, so service could not start up
- `networkd` was managing the `tailscale0` service so Tailscale couldn't enable `MagicDNS`.

## How to use

1. Build the sysext: `./create_tailscale_sysext.sh 1.76.6 tailscale` and `scp` to a running Flatcar machine without the `tailscale` sysext already configured. `ssh` into the Flatcar machine.
2. Ensure `systemctl status tailscaled.service` does not exist
3. `mv /path/to/tailscale.raw /etc/extensions/tailscale.raw`
4. `systemd-sysext refresh`
5. Ensure `systemctl status tailscaled.service` now exists
6. `systemctl start tailscaled.service`
7. Ensure `networkctl list` shows `tailscale0` as `unmanaged` (note, the state will show as `degraded` until you `tailscale up` which isn't needed for this test).

## Testing done

- [x] Ran the above tests
- [x] Ran `flatcar-reset` with an Ignition config that fetched the `tailscale.raw` from an S3 bucket, and checked starting service / checked interface unmanaged / etc...

Closes: https://github.com/flatcar/sysext-bakery/issues/105

### DIFF
--- a/create_tailscale_sysext.sh
+++ b/create_tailscale_sysext.sh
@@ -34,13 +34,16 @@ curl -o "${TMP_DIR}/${TARBALL}" -fsSL "${URL}"
 
 tar xf "${TMP_DIR}/${TARBALL}" -C "${TMP_DIR}" --strip-components=1
 
-mkdir -p "${SYSEXTNAME}"/usr/local/{bin,sbin,lib/{systemd/system,extension-release.d}}
+mkdir -p "${SYSEXTNAME}"/usr/{bin,sbin,lib/{systemd/system,extension-release.d,tmpfiles.d},share/tailscale}
 
-mv "${TMP_DIR}/tailscale" "${SYSEXTNAME}/usr/local/bin/tailscale"
-mv "${TMP_DIR}/tailscaled" "${SYSEXTNAME}/usr/local/sbin/tailscaled"
-mv "${TMP_DIR}/systemd/tailscaled.service" "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+mv "${TMP_DIR}/tailscale" "${SYSEXTNAME}/usr/bin/tailscale"
+mv "${TMP_DIR}/tailscaled" "${SYSEXTNAME}/usr/sbin/tailscaled"
+mv "${TMP_DIR}/systemd/tailscaled.service" "${SYSEXTNAME}/usr/lib/systemd/system/tailscaled.service"
+mv "${TMP_DIR}/systemd/tailscaled.defaults" "${SYSEXTNAME}/usr/share/tailscale/tailscaled.defaults"
 
-sed -i 's/--port.*//g' "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+cat <<EOF >"${SYSEXTNAME}"/usr/lib/tmpfiles.d/10-tailscale.conf
+C /etc/default/tailscaled - - - - /usr/share/tailscale/tailscaled.defaults
+EOF
 
 rm -rf "${TMP_DIR}"
 
--- a/create_tailscale_sysext.sh
+++ b/create_tailscale_sysext.sh
@@ -34,19 +34,12 @@ curl -o "${TMP_DIR}/${TARBALL}" -fsSL "${URL}"
 
 tar xf "${TMP_DIR}/${TARBALL}" -C "${TMP_DIR}" --strip-components=1
 
-mkdir -p "${SYSEXTNAME}"/usr/{bin,sbin,lib/{systemd/system/tailscaled.service.d,systemd/network,extension-release.d,tmpfiles.d},share/tailscale}
+mkdir -p "${SYSEXTNAME}"/usr/{bin,sbin,lib/{systemd/system,systemd/network,extension-release.d,tmpfiles.d},share/tailscale}
 
 mv "${TMP_DIR}/tailscale" "${SYSEXTNAME}/usr/bin/tailscale"
 mv "${TMP_DIR}/tailscaled" "${SYSEXTNAME}/usr/sbin/tailscaled"
 mv "${TMP_DIR}/systemd/tailscaled.service" "${SYSEXTNAME}/usr/lib/systemd/system/tailscaled.service"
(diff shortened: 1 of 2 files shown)


create_tailscale_sysext.sh
```diff
@@ -34,13 +34,32 @@ curl -o "${TMP_DIR}/${TARBALL}" -fsSL "${URL}"
...
+mv "${TMP_DIR}/systemd/tailscaled.service" "${SYSEXTNAME}/usr/lib/systemd/system/tailscaled.service"
+mv "${TMP_DIR}/systemd/tailscaled.defaults" "${SYSEXTNAME}/usr/share/tailscale/tailscaled.defaults"
 
-sed -i 's/--port.*//g' "${SYSEXTNAME}/usr/local/lib/systemd/system/tailscaled.service"
+cat <<EOF >"${SYSEXTNAME}"/usr/lib/systemd/system/tailscaled.service.d/10-networkd-reload.conf
+# Reload systemd-networkd.service to pick up 50-tailscale.network
+
+[Service]
+ExecStartPre=systemctl reload systemd-networkd.service
+EOF
```
jmacdonagh, line 49: Is there a better way to achieve this? The sysext will place `/usr/lib/systemd/network/50-tailscale.network` but that happens after `systemd-networkd.service` has started. It would be better if we had some mechanism to reload this service when this `sysext` is loaded.

jmacdonagh: When you run `networkctl list` what does it show for “TYPE” for `tailscale0`?

Mine was showing “none” but I expected it to show “tun” or “tunnel”. If it did, then a default `networkctl` config would have made it unmanaged

jmacdonagh: Ah nevermind, did a clean `flatcar-reset` with a newly build sysext and it's working as you described. I'm almost certain I tried this before but maybe not.

jmacdonagh: Ok, there's some kind of race condition with my (minimal) Ignition config.

I observed what you have, with a minimal Ignition config, the link comes up umanaged.

However, if I add just a bit more config to Ignition (specifically, mount another volume to `/var/lib/docker`), the link repeatedly comes up as managed.

```yaml
variant: flatcar
version: 1.0.0
storage:
  filesystems:
    - device: /dev/disk/by-id/scsi-0HC_Volume_XXXXXX
      format: ext4
      wipe_filesystem: false
      label: VOLUME
  files:
    - path: /opt/extensions/tailscale/tailscale-1.76.6-x86-64.raw
      contents:
        source: https://XXXXXX.s3.com/tailscale.raw
    - path: /etc/sysupdate.d/noop.conf
      contents:
        source: https://github.com/flatcar/sysext-bakery/releases/download/latest/noop.conf
    - path: /etc/sysupdate.tailscale.d/tailscale.conf
      contents:
        source: https://github.com/flatcar/sysext-bakery/releases/download/latest/tailscale.conf
  links:
    - path: /etc/resolv.conf
      target: /run/systemd/resolve/stub-resolv.conf
      overwrite: true
    - path: /etc/extensions/tailscale.raw
      target: /opt/extensions/tailscale/tailscale-1.76.6-x86-64.raw
      hard: false
    - path: /etc/systemd/system/multi-user.target.wants/tailscaled.service
      target: /usr/local/lib/systemd/system/tailscaled.service
      overwrite: true
systemd:
  units:
    # Docker volume mount
    - name: var-lib-docker.mount
      enabled: true
      contents: |
        [Unit]
        Description=Mount external volume to /var/lib/docker
        Before=local-fs.target
        [Mount]
        What=/dev/disk/by-label/VOLUME
        Where=/var/lib/docker
        Type=ext4
        [Install]
        WantedBy=local-fs.target
    - name: docker.service
      dropins:
        - name: 10-wait-docker.conf
          contents: |
            [Unit]
            After=var-lib-docker.mount
            Requires=var-lib-docker.mount
    # Tailscale sysext
    - name: systemd-sysupdate.timer
      enabled: true
    - name: systemd-sysupdate.service
      dropins:
        - name: tailscale.conf
          contents: |
            [Service]
            ExecStartPre=/usr/lib/systemd/systemd-sysupdate -C tailscale update
        - name: sysext.conf
          contents: |
            [Service]
            ExecStartPost=systemctl restart systemd-sysext
```

There's no explicit link between the docker mount and sysext, so I'm not sure what's going on.

jmacdonagh: Agreed. There is no need to reload `systemd-networkd`, a reload works perfectly and avoids bringing `eth0` down and up.

I've added a PR to do exactly this: https://github.com/flatcar/init/pull/128

Full disclosure, I've moved to just having `tailscale` run in docker instead of using this sysext, but happy to test/fix this for this PR ;)

jmacdonagh: Unfortunately, you lose out on some optional tailscale functionality when running in Docker, such as built-in SSH and its file transfer features. These work, but since you're in the equivalent of a `chroot` it's not as useful as it running natively. I don't use those features so Docker works for me. In fact, it's an officially supported way of running Tailscale: https://tailscale.com/kb/1282/docker

All this being said, do you want another thing to maintain and update? 😄 

I suppose another solution would be to build sysexts directly from Gentoo: https://packages.gentoo.org/packages/net-vpn/tailscale

But that'll quickly get tricky maintaining runtime dependencies.

t-lo:
I can reproduce the breakage on 4152.2.0 - repros in qemu for me all the time.
Symptoms include inability to SSH into the qemu machine after the sysext was merged.

This is caused by a "unified `/usr/bin/`"; `/usr/sbin` has become a symlink to `/usr/bin/`. So when the sysext is merged, all of `/usr/sbin` disappears as it is turned from a symlink into a real directory with only one file in it - `tailscaled`.

I think we should just put everything into `/usr/bin/` and update the service unit file accordingly. I'm currently working on a re-write of the sysext scripting to make it more usable and to reduce duplicate code; I'll factor @jmacdonagh 's changes in and attribute you as co-author.

jmacdonagh:
Thanks @t-lo! I will close this PR in anticipation that https://github.com/flatcar/sysext-bakery/pull/115 will be merged in!

PR #108, opened 2025-01-28
README, docs/: overhaul the bakery documentation

t-lo:
This change is a thorough documentation overhaul.

The main README has been refactored to provide a top-down, user-centric view of the Bakery. All specific sysexts are discussed in separate files. All sysexts now come with full-blown ready-to-use example configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,494 +1,308 @@
-# sysext-bakery: Recipes for baking systemd-sysext images
+# Ready-to use System Extensions for Flatcar (and other distros)
 
-[Systemd-sysext images](https://www.freedesktop.org/software/systemd/man/systemd-sysext.html) are overlay images for `/usr`, allowing to extend the base OS with custom (static) binaries.
-Flatcar Container Linux as an OS without a package manager is a good fit for extension through systemd-sysext.
-The tools in this repository help you to create your own sysext images bundeling software to extend your base OS.
-The current focus is on Docker and containerd, contributions are welcome for other software.
-See the section at the end on how to bundle any software with the Flix and Flatwrap tools.
+The Sysext Bakery serves 3 main functions:
 
-## Systemd-sysext
-
-The `NAME.raw` sysext images (or `NAME` sysext directories) can be placed under `/etc/extensions/` or `/var/lib/extensions` to be activated on boot by `systemd-sysext.service`.
-Images can specify to require systemd to do a daemon reload (needs systemd 255, Flatcar ships `ensure-sysext.service` as workaround to automatically load the image's services).
-
-A current limitation of `systemd-sysext` is that you need to use `TARGET.upholds` symlinks (supported from systemd 254, similar to `.wants`) or `Upholds=` drop-ins for the target units to start your units.
-For current versions of Flatcar (systemd 252) you need to use `systemctl restart systemd-sysext ensure-sysext` to reload the sysext images and start the services and a manual `systemd-sysext refresh` is not recommended.
-
-The compatibility mechanism of sysext images requires a metadata file in the image under `usr/lib/extension-release.d/extension-release.NAME`.
-It needs to contain a matching OS `ID`, and either a matching `VERSION_ID` or `SYSEXT_LEVEL`. Here you can also set `EXTENSION_RELOAD_MANAGER=1` for a systemd daemon reload.
-Since the rapid release cycle and automatic updates of Flatcar Container Linux make it hard to rely on particular OS libraries by specifying a dependency of the sysext image to the OS version, it is not recommended to match by `VERSION_ID`.
-Instead, Flatcar defined the `SYSEXT_LEVEL` value `1.0` to match for.
-You can also use `ID=_any` and then neither `SYSEXT_LEVEL` nor `VERSION_ID` are needed.
-The sysext image should only include static binaries.
-
-Inside the image, binaries should be placed under `usr/bin/` and systemd units under `usr/lib/systemd/system/`.
-While placing symlinks in the image itself to enable the units in the same way as systemd would normally do (like `sockets.target.wants/my.socket` → `../my.socket`) is still currently supported, this is not a recommended practice.
-The recommended way is to ship drop-ins for the target units that start your unit.
-The drop-in file should use the `Upholds=` property in the `[Unit]` section.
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -1,494 +1,301 @@
...
-While placing symlinks in the image itself to enable the units in the same way as systemd would normally do (like `sockets.target.wants/my.socket` → `../my.socket`) is still currently supported, this is not a recommended practice.
-The recommended way is to ship drop-ins for the target units that start your unit.
-The drop-in file should use the `Upholds=` property in the `[Unit]` section.
-For example, for starting `docker.socket` we would use a drop-in for `sockets.target` placed in `usr/lib/systemd/system/sockets.target.d/10-docker-socket.conf` with the following contents:
 
-```
-[Unit]
-Upholds=docker.socket
-```
+## What's a sysext and how does it extend Flatcar?
```
brendank310, line 39: Putting this above the table seems more natural to me. It answers the "what is the project and what it can do" early

README.md
```diff
@@ -1,494 +1,301 @@
...
-          contents: |
-            [Service]
-            ExecStartPost=systemctl restart systemd-sysext
 ```
 
-In the [Flatcar docs](https://www.flatcar.org/docs/latest/provisioning/sysext/) you can find an Ignition configuration that explicitly sets the update configurations instead of downloading them.
+## Extension auto-updates
 
-The updates works by [`systemd-sysupdate`](https://www.freedesktop.org/software/systemd/man/sysupdate.d.html) fetching the `SHA256SUMS` file of the generated artifacts, which holds the list of built images with their respective SHA256 digest.
+Using the flexible approach above we can instruct `systemd-sysupdate` to poll the bakery for updates.
```
mkulke, line 104: maybe indicate when polling occurs (hourly, daily, at reboot)?

README.md
```diff
@@ -1,494 +1,301 @@
...
+Note that this snippet inlines all of the sysupdate configuration for documentation purposes.
+Most extensions published in the Bakery ship sysupdate configuration as part of the releases, so these can be downloaded by Ignition instead of being inlined.
+(And consequently, that's what most of the configuration examples of individual extensions do).
+Check out the respective extensions' readme for details.
 
-1. Provide a different configuration to setup the nats-server to act as a leaf node to a pre-existing wasmCloud deployment (`/etc/nats-server.conf`).
-2. Provide a set of credentials for the nats-server leaf node to connect with (`/etc/nats.creds`).
-3. Override the bundled `NATS_CONFIG` environment variable to point it to the newly created configuration (`NATS_CONFIG=/etc/nats-server.conf`).
-4. Override the lattice the wasmCloud host is configured to connect (`WASMCLOUD_LATTICE=<redacted>`).
+Also, we include a dummy `noop.conf` for systemd-sysupdate to work around a spurious error message.
```
mkulke, line 225: it's unclear whether the user would need to add a similar clause/only required temporarily/only for the sake of this example. maybe add a remark to provide context?

README.md
```diff
@@ -514,74 +321,53 @@ sudo systemd-sysext refresh
...
-  myext.conf
-./bake_flatcar_image.sh --fetch --vendor qemu_uefi myext:myext-1.0.1-x86-64.raw
-```
-
-The script supports all vendors and clouds natively supported by Flatcar.
-
-### Flix and Flatwrap
+## Flix and Flatwrap
 
 The Flix and Flatwrap tools both convert a given chroot folder into a systemd-sysext image.
```
mkulke, line 362: It's not entirely clear to me what "chroot folder" implies. I presume it's a folder that I would chroot into (a chroot-able folder?) If that's the case, the above instructions indicate that only files ina `/usr` hiararchy are considered valid. this would be a contradiction?

README.md
```diff
@@ -514,74 +321,53 @@ sudo systemd-sysext refresh
...
+E.g. [create_ollama_sysext.sh](https://github.com/flatcar/sysext-bakery/blob/main/create_ollama_sysext.sh) is a pretty simple though full-featured one.
 
-```
-OS=_any ARCH=arm64 ./create_docker_sysext.sh 24.0.9 mydocker
-[… writes mydocker.raw into current directory …]
-```
+Create a new script `create_EXTNAME_sysext.sh` which
+1. Creates a subdirectory `EXTNAME` which will later be used to generate the sysext file system from.
+2. Download a release of the application you want to ship and store them in suitable subdirectories below `EXTNAME/`.
+   Make sure the application is self-contained **and** doesn't ship any system libraries like glibc etc.
```
mkulke, line 342:
```suggestion
   Make sure the application is self-contained **and** doesn't ship any system libraries like glibc etc. If you need such a library in a sysext, please consult the Flix/Flatwrap instructions below.
```

README.md
```diff
@@ -514,74 +321,53 @@ sudo systemd-sysext refresh
...
 The Flix and Flatwrap tools both convert a given chroot folder into a systemd-sysext image.
 You have to specify which files should be made available to the host.
 
 The Flix tool rewrites specified binaries to use a custom library path.
 You also have to specify needed resource folders and you can specify systemd units, too.
 
+These tools can be used to safely ship library dependencies of OS libraries (like e.g. glibc) in a custom path.
+The custom path ("sysroot") avoids clashes with Flatcar's native OS libraries.
+
 Here examples with Flix:
```
mkulke, line 371:
```suggestion
Flix examples:
```

README.md
t-lo: I've never _seen_ the issue but have carried along this workaround for the time being. Kai added this some time ago, might be cargo cult :shrug:

README.md
t-lo: That's a good point, will restructure.

README.md
t-lo: Good point, done.

README.md
t-lo: It's actually just "root" folder; updated.

mkulke (approved):

t-lo:
> Some minor nits on the README, but looks good!

Thank you very much @skoeva , merged!

t-lo:
I'm going to merge this so we have it up in time for FOSDEM. Should amendments etc. be necessary we'll file follow-up PRs.

PR #109, opened 2025-02-02
Add Nebula sysext

SleepyLeslie:
# [Title: describe the change in one sentence]

This PR adds a [Nebula](https://github.com/slackhq/nebula) sysext, including the build script and documentation.

## How to use

Run `create_nebula_sysext.sh 1.9.5 nebula` to produce the squashfs image `nebula.raw`. Verify that it contains `usr/bin/nebula` and `/usr/bin/nebula-cert` and both should run.

## Testing done

I ran the CI pipeline in my fork and successfully produced correct squashfs images for amd64 and arm64, as well as the `.conf` file for sysupdate.

### DIFF
--- a/docs/nebula.md
+++ b/docs/nebula.md
@@ -1,0 +1,45 @@
+# Nebula sysext
+
+This sysext ships [Nebula](https://github.com/slackhq/nebula).
+
+## Usage
+
+Refer to the following Butane snippet that enables Nebula v1.9.5 for an x86-64 machine with automated updates using `systemd-sysupdate`.
+
+```yaml
+variant: flatcar
+version: 1.0.0
+
+storage:
+  files:
+    - path: /opt/extensions/nebula/nebula-v1.9.5-x86-64.raw
+      mode: 0644
+      contents:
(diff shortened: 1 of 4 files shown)


docs/nebula.md
```diff
@@ -0,0 +1,45 @@
+# Nebula sysext
+
+This sysext ships [Nebula](https://github.com/slackhq/nebula).
+
+## Usage
+
+Refer to the following Butane snippet that enables Nebula v1.9.5 for an x86-64 machine with automated updates using `systemd-sysupdate`.
+
```
t-lo, line 8: Can we add a note here that the user would need to supply a custom [Nebula config](https://github.com/slackhq/nebula/blob/master/examples/config.yml) which they could do by embedding it into the Butane configuration?

t-lo:
Nice work @SleepyLeslie , thank you! Looks good in general, one minor nit regarding documentation.

I've tested the build and Butane config snippet locally, everything works as expected.

docs/nebula.md
SleepyLeslie: Done!

PR #111, opened 2025-02-04
pages: host static documentation

tormath1:
In this PR, we host the documentation from `docs/` on `flatcar.github.io`.

Once merged, we just need to enable the `pages` feature of this repository to enable the hosting on `https://flatcar.github.io/sysext-bakery`

As the sysext-bakery is not strictly bound to Flatcar, I think it makes sense to have its own documentation.

This commit is currently deployed on my fork: https://tormath1.github.io/sysext-bakery/

NOTE: To avoid duplication of the README, I added a symlink but it means that we will have this:
```md
---
nav_order: 1
title: Overview
---
```

on top of the `README.md` on the GitHub repository.


t-lo (approved):
LGTM, thanks for the quick turnaround! I particularly like the README smylink so we don't lose the landing page in the github repo.

We might even want to include a link to flatcar.github-io/sysext-bakery in the index file / README (even if it would be self-referential in the hosted pages) so people can click it in the Repo landing page.

PR #126, opened 2025-03-26
Allow creating sysext with 'latest' version

reavessm:
# Allow for building sysext without knowing the version

This allows you to run `./bakery.sh create <sysext> latest`.

## How to use

1. before patch
2. `./bakery.sh create k3s latest`
3. apply patch
4. `./bakery.sh create k3s latest`

## Testing done

Above steps

### DIFF
--- a/bakery.sh
+++ b/bakery.sh
@@ -116,6 +116,10 @@ function create_sysext() {
     return 1
   fi
 
+  if [[ ${version} == latest ]] ; then
+    version="$(list_sysext "$extname" --latest true)"
+  fi
+
   local workdir="$(mktemp -d)"
   local sysextroot_tmp="$(mktemp -d)"
   trap "rm -rf '${workdir}' '${sysextroot_tmp}'" EXIT


t-lo:
Thank you for the PR!

"latest" can return multiple versions - the Kubernetes sysext for instance does that to list latest releases of all supported major versions (see https://github.com/flatcar/sysext-bakery/blob/main/kubernetes.sysext/create.sh#L11).

If we want to support "latest" in `bakery.sh` then we should put everything below into a loop like the release dispatcher currently does it: https://github.com/flatcar/sysext-bakery/blob/main/release_dispatcher.sh#L34 (and preferably into its own sub-function for readability).

reavessm:
hmm that does change things a bit.  Would it be better to build EACH of the multiple releases or only the absolute latest of the "latest"?  I originally envisioned this feature as quick way to check your sysext works so I'm partial to only building the latest of latest, but I'm open to either

t-lo:
I'm trying to prevent introducing two different concepts of "latest": the current one via `./bakery.sh list <extension> --latest true` which may produce multiple "latests" depending on the sysext to be built, and a new one where only the highest release version's latest is being built.

The reason for the current concept is that more complex upstreams that have widespread adoption, like Kubernetes and k3s, support multiple major releases in parallel. Kubernetes for instance releases new versions for major release series 1.29, 1.30, 1.31 and 1.32, as upgrades across major release series can be painful for users and operators. We want to reflect that in the bakery releases and not force users to immediately switch to a new major version when it is first released, but instead receive further updates for the major series they are currently on - for as long as upstream maintains these.

reavessm:
Fair enough.  I hadn't considered that kubernetes has multiple "latest"s.  I think we can just close this then

PR #130, opened 2025-03-27
Docker sysext improvement

darkspadez:
# Improve the docker sysext

There is now a containerd sysext removing the need to use docker sysext to build just containerd/runc. Also defaults to always build docker ( it is the docker sysext ) and remove containerd in favor of containerd sysext


t-lo:
Thank you for your contribution! Agree that removing `--without docker` makes a lot of sense. I'd however retain the option to include containerd, even as the default. It makes the sysext self-contained, which is preferred - sysexts are not supposed to replace traditional package management, and dependency handling largely does not exist.
Keeping containerd in by default provides a soft path forward for folks to migrate to a composed set-up.
Also, may I bug you to please update the documentation at `docs/docker.md`?

darkspadez:
Sure no problem, I will go ahead and tweak the PR to bring back containerd/runc by default. I plan on updating docs on this PR once we finalize the code aspects of it. I have other updated for docs too that I will do with another pull request.

Question though if docker is built w/ containerd/runc is there a way to overwrite it with a containerd sysext? When I last tested it, it wasnt possible but unsure with the new build system if it is or not.

Currently if you want to run the "community" docker sysext you need to /dev/null flatcars default docker + containerd

```
    - path: /etc/extensions/docker-flatcar.raw
      target: /dev/null
      overwrite: true
    - path: /etc/extensions/containerd-flatcar.raw
      target: /dev/null
      overwrite: true
 ```

This is where i saw a discrepancy with the community one compared to internal one and was trying to mimic it here.

I am all for including containerd inside docker, just need to know if you currently can overwrite it someway with containerd sysext or if possible to add that functionality.

**Comment by darkspadez** [comment]
Actually thinking about it slightly more after going through more docs, will docker still not work with the built-in containerd?

Currently for nerdctl we have this:
```
Nerdctl sysext
This sysext ships [nerdctl](https://github.com/containerd/nerdctl). It requires containerd, either built-in or via a [containerd](https://flatcar.github.io/docs/containerd.md) or [docker](https://flatcar.github.io/docs/docker.md) sysext.
```

Allow docker sysext to either use built-in containerd which would allow it to work by default, or with containerd sysext.

darkspadez:
@t-lo I have been doing some testing and here are my results

When using the sysexts docker and containerd from this repo, containerd gets overwritten from the docker sysext. This makes containerd sysext pointless for when running with docker. I narrowed this down to sysext loading extensions alphabetically.

When using docker built w/o containerd and with either the containerd sysext from this repo or the default containerd from flatcar by default, it all works without issue.

I think best option would be to remove containerd requirement and either have them use containerd sysext from this repo or default one included from flatcar. This will allow for more robust setups.

PR #132, opened 2025-03-29
keepalived: do not use /usr/sbin

### LINKED ISSUE #131: Keepalived sysext overwrites Flatcar's `sbin/` symlink
## Description

Running the Keepalived sysext overwrites Flatcar's `/usr/sbin` symlink causing systemic failure.

Same issue as #105 

## Impact

Unable to boot a stable Flatcar instance with the Keepalived sysext installed.

## Environment and steps to reproduce

1. **Set-up**:
  - Create a minimal Flatcar instance with the following config:
  ```
  variant: flatcar
  version: 1.1.0
  
  storage:
    files:
      - path: /etc/extensions/keepalived.raw
         contents:
         source: https://extensions.flatcar.org/extensions/keepalived-v2.3.1-x86-64.raw 
  ```
3. **Task**:
    - Boot system (local via QEMU is fine)
5. **Action(s)**:
    - Boot system (local via QEMU is fine)
4. **Error**:
    - `[FAILED] Failed to start update-engine.service - Update Engine.`
    - `Update Strategy: No Reboot` displayed on console shell
    - Remote SSH connections dropped
    - etc.

## Expected behavior

Successful boot.

## Additional information

Please add any information here that does not fit the above format.

t-lo:
/usr/sbin is a symlink to /usr/bin on recent Flatcar releases. Sysexts shipping /usr/sbin would overwrite that symlink and sbin would contain only the contents that sysext ships.

We copy keepalived to /usr/bin and also pin the Alpine Linux release used for the build.

After merge we need to delete the keepalived releases to force a rebuild.

Fixes #131

### DIFF
--- a/keepalived.sysext/create.sh
+++ b/keepalived.sysext/create.sh
@@ -30,9 +30,13 @@ function populate_sysext_root() {
               -i \
               -v "$(pwd)":/install_root \
               --platform "linux/${img_arch}" \
-              alpine \
+              --pull always \
+              alpine:v3.21 \
                   /install_root/build.sh "${version}" "$user_group"
 
+  # /usr/sbin is a symlink to /usr/bin on Flatcar.
+  mv usr/sbin/keepalived usr/bin
+  rmdir usr/sbin
   cp -aR usr "${sysextroot}"/
 }
 # --


ader1990 (approved):
LGTM

chewi (approved):
This is fine, but I wonder if we can avoid this more generally by populating sysextroot with symlinks in the first place?

t-lo:
> This is fine, but I wonder if we can avoid this more generally by populating sysextroot with symlinks in the first place?

@chewi Yes, that's definitely something we should investigate. Like, force-abort a build with an error message in `lib/generate.sh` when we detect `/usr/sbin` to be present.

Issue #123, opened 2025-03-26
Correct Ignition URL for Kubernetes?


asgeirn:
Looking at https://github.com/flatcar/sysext-bakery/blob/02939cc7e21191a534cf865b831f806cd5a8e56d/docs/kubernetes.md?plain=1#L34 the correct URL for Ignition should be https://extensions.flatcar.org/extensions/kubernetes-v1.31.conf

However, looking at https://github.com/flatcar/sysext-bakery/blob/02939cc7e21191a534cf865b831f806cd5a8e56d/tools/http-url-rewrite-server/Caddyfile#L34 the correct URL should either be `<extension>-<version>.conf` or `<extension>.conf`

Since releases are either individual releases or extension metadata releases, there is no way of obtaining the configuration file for a Kubernetes release.

URLs to test:

- https://extensions.flatcar.org/extensions/kubernetes-v1.32.conf rewrites to https://github.com/flatcar/sysext-bakery/releases/download/kubernetes-v1.32/kubernetes-v1.32.conf which does not exist.
- https://extensions.flatcar.org/extensions/kubernetes.conf rewrites to https://github.com/flatcar/sysext-bakery/releases/download/kubernetes/kubernetes.conf which exists
- https://extensions.flatcar.org/extensions/kubernetes-v1.32.2.conf rewrites to https://github.com/flatcar/sysext-bakery/releases/download/kubernetes-v1.32.2/kubernetes-v1.32.2.conf which does not exist

t-lo:
Nice catch!
Kubernetes is a bit special, we missed to handle that in the URL rewriter. The server has been updated and I'll file a PR here momentarily.

Please follow the pattern `https://extensions.flatcar.org/extensions/kubernetes/kubernetes-<version>.conf` .

t-lo:
@asgeirn Can you check again? I filed https://github.com/flatcar/sysext-bakery/pull/124 and updated the redirect server.

asgeirn:
Thanks!  The URL https://extensions.flatcar.org/extensions/kubernetes/kubernetes-v1.32.conf now correctly resolves to https://github.com/flatcar/sysext-bakery/releases/download/kubernetes/kubernetes-v1.32.conf

However, for the raw file this Ignition configuration specifies:

    Path=https://extensions.flatcar.org/extensions/kubernetes/
    MatchPattern=kubernetes-v1.32.@v-%a.raw

[...]

Issue #127, opened 2025-03-26
docker-compose RAW image URL reports 404


genzj:
`systemd-sysupdate` failed and seeing following log in `journalctl -xe -u systemd-sysupdate.service`

```
Mar 26 19:59:37 localhost systemd-sysupdate[617743]: \ Acquiring https://extensions.flatcar.org/extensions/docker-compose/docker-compose-2.34.0-x86-64.raw → /opt/extensions/docker-compose/docker-compose-2.34.0-x86-64.raw...
Mar 26 19:59:37 localhost systemd-sysupdate[617745]: Pulling 'https://extensions.flatcar.org/extensions/docker-compose/docker-compose-2.34.0-x86-64.raw', saving as '/opt/extensions/docker-compose/.#sysupdatedocker-compose-2.34.0-x86-64.raw8547f5c30572d594'.
Mar 26 19:59:38 localhost systemd-sysupdate[617745]: HTTP request to https://extensions.flatcar.org/extensions/docker-compose/docker-compose-2.34.0-x86-64.raw failed with code 404.
Mar 26 19:59:38 localhost systemd-sysupdate[617745]: Failed to retrieve image file.
```

A working URL is `https://extensions.flatcar.org/extensions/docker-compose-2.34.0-x86-64.raw`, which does not include the `docker-compose/` portion before the image name.

It appears to be related to the Caddy file change in #124. The pulling URL matches the following rule and is redirected to `https://github.com/flatcar/sysext-bakery/releases/download/docker-compose/docker-compose-2.34.0-x86-64.raw`, but it lacks the version in `/docker-compose/` portion.

https://github.com/flatcar/sysext-bakery/blob/510b5193ce659392efa51857b5e03e9b250f0287/tools/http-url-rewrite-server/Caddyfile#L39-L44

The expected redir target should be `https://github.com/flatcar/sysext-bakery/releases/download/docker-compose-2.34.0/docker-compose-2.34.0-x86-64.raw`

t-lo:
Thank you for the report! I've filed https://github.com/flatcar/sysext-bakery/pull/133 to fix it.

t-lo:
I've updated the live redirect server, the URLs should work now.

genzj:
Many thanks for addressing the issue!

PR #134, opened 2025-03-30
Caddyfile: support subpaths only for sysupdate confs

t-lo:
This change fixes an issue introduced by
510b5193ce659392efa51857b5e03e9b250f0287
when introducing sub-path retaining for confs and images. Images aren't really available from these sub-paths so we need to limit that feature to sysupdate confs only.

Fixes https://github.com/flatcar/sysext-bakery/issues/123 , #127 .


t-lo:
Updated the live redirect server with the change so it can be tested,

chewi (approved):

PR #135, opened 2025-03-30
[FEATURE] Add support for custom bakery url

darkspadez:
# [FEATURE] Add support for custom bakery url

Currently hardcoded to extensions.flatcar.org/extensions/, though forks and individual bakerys would need to edit files directly to use their own.

## How to use

Add `--bakery <url>` to `bakery.sh` during build time or `--bakery <url>` to `tools/bake_flatcar_image.sh `. This makes it easier for forks and other "bakerys" to specify their url instead of editing the files directly.

### DIFF
--- a/lib/libbakery.sh
+++ b/lib/libbakery.sh
@@ -11,7 +11,7 @@ libroot="$(dirname "${BASH_SOURCE[0]}")"
 scriptroot="$(cd "$(dirname "${BASH_SOURCE[0]}")/../"; pwd)"
 
 bakery="flatcar/sysext-bakery"
-bakery_hub="extensions.flatcar.org"
+bakery_base_url="https://extensions.flatcar.org/extensions"
 
 # Add new library function scripts here:
 source "${libroot}/helpers.sh"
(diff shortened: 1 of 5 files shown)


lib/libbakery.sh
```diff
@@ -11,7 +11,7 @@ libroot="$(dirname "${BASH_SOURCE[0]}")"
 scriptroot="$(cd "$(dirname "${BASH_SOURCE[0]}")/../"; pwd)"
 
 bakery="flatcar/sysext-bakery"
```
t-lo, line 13: We need to take care of this one as well, it is used to query version information of existing images from github to determine which ones to build.

t-lo:
Thank you for your contribution! May I suggest putting both `bakery=` and `bakery_url`  into a separate env file which is sourced by `lib/libbakery` if it exists, to overwrite the defaults? This would make a fork work more flexibly as there only needs to be one change on top (the custom env file) - which is easy to rebase.

darkspadez:
@t-lo Just to confirm, switch from --bakery argument to an environment file?

t-lo:
> @t-lo Just to confirm, switch from --bakery argument to an environment file?

@darkspadez Yes, an optional one. I.e. leave defaults in `lib/libbakery.sh` as is but source a custom env file after that if that env file exists.

This should make it easier for forks to customise the builds.

PR #138, opened 2025-04-01
Keepalived: Fix used alpine image

UnholyRope:
Fixes the keepalived extension to be built using a valid alpine image.

### DIFF
--- a/keepalived.sysext/create.sh
+++ b/keepalived.sysext/create.sh
@@ -19,7 +19,7 @@ function populate_sysext_root() {
   local img_arch="$(arch_transform 'x86-64' 'amd64' "$arch")"
   img_arch="$(arch_transform 'arm64' 'arm64/v8' "$img_arch")"
 
-  local image="docker.io/${img_arch}/alpine:3.19"
+  local image="docker.io/${img_arch}/alpine:3.21"
 
   announce "Building keepalived $version for $arch"
 
@@ -31,7 +31,7 @@ function populate_sysext_root() {
               -v "$(pwd)":/install_root \
               --platform "linux/${img_arch}" \
               --pull always \
-              alpine:v3.21 \
+              ${image} \
                   /install_root/build.sh "${version}" "$user_group"
 
   # /usr/sbin is a symlink to /usr/bin on Flatcar.


keepalived.sysext/create.sh
```diff
@@ -19,7 +19,7 @@ function populate_sysext_root() {
   local img_arch="$(arch_transform 'x86-64' 'amd64' "$arch")"
   img_arch="$(arch_transform 'arm64' 'arm64/v8' "$img_arch")"
 
-  local image="docker.io/${img_arch}/alpine:3.19"
+  local image="docker.io/${img_arch}/alpine:3.21"
```
t-lo, line 22:
```suggestion
  local image="docker.io/alpine:3.21"
```
as we explicitly specify `--platform` on the `docker run`  command line. For some reason the original version even workd for x86-64 but breaks for arm64.

UnholyRope: Thanks for catching that. I updated it accordingly.

UnholyRope:
> Thank you for your fix! This issue must have snuck in during the previous `/usr/sbin` fix. One minor nit to make arm64 builds work (see comment), otherwise the PR is good to go.
> 
> BTW you can test arm64 builds on non-native architectures, e.g. x86-64, if you have binfmt set up on the build host. Our bakery CI works that way.
> 
> If you like you can even re-add "keeplived latest" `release_build_versions.txt` as the most recent keepalived release, 2.3.3., fixes a build issue on Alpine that reproducibly broke the 2.3.2 build.

I added `keepalived latest` back the `release_buld_versions.txt` and ran the pipeline. The build itself works but the release body is too long:
```
Error: Validation Failed: {"resource":"Release","code":"custom","field":"body","message":"body is too long (maximum is 125000 characters)"} - https://docs.github.com/rest/releases/releases#create-a-release
```

t-lo:
Ah, okay. Looks like we should capture the build log in a file and make it part of the release artefacts instead of the release text.

UnholyRope:
I removed the build output from the release as it can be viewed in the pipeline if needed.
The extension releases now look like this: [https://github.com/UnholyRope/sysext-bakery/releases/tag/keepalived-v2.3.3](https://github.com/UnholyRope/sysext-bakery/releases/tag/keepalived-v2.3.3).

t-lo (approved):
LGTM, tested the builds locally. Thank you @UnholyRope !

t-lo:
Merging, thanks for the good work! Will follow up with a PR to CI release automation to capture build output into a separate log file.

PR #143, opened 2025-04-09
k3s,kubernetes,rke2: fix sysupdate symlink names

### LINKED ISSUE #140: K3s: Patchlevel sysupdate doesn't work correctly
## Description

The transfer definition file defines `CurrentSymlink=/etc/extensions/k3s-v1.32.raw`, while the the docs refer to `/etc/extensions/k3s.raw` ([links](https://github.com/flatcar/sysext-bakery/blob/d548304e38ed7da8fe37ba436b988dc4005ce6a4/docs/k3s.md?plain=1#L41-L44) and [sysupdate dropin](https://github.com/flatcar/sysext-bakery/blob/d548304e38ed7da8fe37ba436b988dc4005ce6a4/docs/k3s.md?plain=1#L49-L57)).

## Impact

Following the docs the sysupdate service will create the symlinked file `/etc/extensions/k3s-v1.32.raw` and cause `systemd-sysext.service` to fail after a reboot.

## Environment and steps to reproduce

1. **Set-up**: Flatcar was installed with the k3s extension as described in the docs. The only difference is that the transfer definition file was put under `/etc/sysupdate.k3s-v1.32.d/k3s-v1.32.conf`.
2. **Task**: Update k3s to the latest minor release.
3. **Action(s)**:
  a. Execute `systemctl start systemd-sysupdate`
  b. Reboot
4. **Error**: `systemd-sysext.service` fails to start.

## Expected behavior

`systemd-sysext.service` should start without issue and load the K3s extension properly.

## Additional information

I tried updating the links and sysupdate unit, but to no avail. Replacing `CurrentSymlink=/etc/extensions/k3s-v1.32.raw` with `CurrentSymlink=/etc/extensions/k3s.raw` in the `k3s-v1.32.conf` (and deleting `/etc/extensions/k3s-v1.32.raw`) seems to fix the issue.
Deleting the updated image (`/opt/extensions/k3s/k3s-v1.32.3+k3s1-x86-64.raw`), executing `systemctl start systemd-sysupdate` again and rebooting updates the extension properly and starts without issue.

t-lo:
We publish k3s, kubernetes, and rke2 with special sysupdate configs to ensure only patch levels are updated. However, we still use <extension> w/o major version as the sysext's name, and we use a matching symlink name.

The special sysupdate configs however use
<extension>-<majorversion>.raw.
This change updates sysupdate configuration generation for the 3 extensions to use the correct symlink.

Fixes #140.

I also added releases v1.32.1 to all 3 sysexts' release_build_versions.txt so we can test this.


t-lo:
# TODO after merge
Delete kubernetes, rke2, and k3s releases and do a full rebuild.

chewi (approved):

t-lo:
Deleted Kubernetes, RKE2, and K3S releases and triggered new CI build workflow to refresh releases: https://github.com/flatcar/sysext-bakery/actions/runs/14378029070

PR #145, opened 2025-04-10
[FEATURE] Env file

darkspadez:
# .env file support

Currently bakery_hub and bakery are hard coded requiring manual intervention when forking. This allows for a .env file to override those values for better fork support.

I use `-s` as this will only source if file is there and is not empty.

## How to use

Move .env.example to .env and update bakery and bakery_hub values and build.

This all stems from https://github.com/flatcar/sysext-bakery/pull/135

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,2 @@
+bakery="flatcar/sysext-bakery"
+bakery_hub="extensions.flatcar.org"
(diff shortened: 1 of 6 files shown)


darkspadez:
@t-lo Here is environment file support!

darkspadez:
@t-lo Done!

t-lo (approved):
LGTM, thank you!

PR #146, opened 2025-04-10
fix(docs): explicitly enable markdown rendering inside html block on index.md

RiRa12621:
# Fix markdown Rendering on docs index

Currently looks like this:
Screenshot 2025-04-10 at 22 30 17

This _should_ do in accordance with https://github.com/planetjekyll/quickrefs/blob/master/FAQ.md#q-jekyll-doesnt-render-markdown-when-adding-html-tags


John15321:
Thanks for the fix!

PR #149, opened 2025-04-17
Containerd improvement

darkspadez:
# Containerd Improvements

Currently containerd is included in the sysexts of docker, k3s, etc. This will remove containerd from them directly and create a new sysext for each like: containerd-<sysext>

This allows us to still provide containerd that comes with docker, k3s, etc while allowing it to be switched out with that of upstream or different ones without causing any issues.

It will automatically build the extension during build time of the upstream sysext of docker, k3s, etc.

I really think we should do something to better handle containerd and I think this is a good compromise.  I will add k3s and any others to this PR but want to make sure this is the route we want to go before spending all the time doing that and updating docs as well.

## Testing done

Included in this pull request is test-build.sh that I used to test building that I will remove. This will need more testing especially w/ the github build system to make sure everything is built correctly.


darkspadez:
This is a replacement/improvement of the following PR https://github.com/flatcar/sysext-bakery/pull/130

PR #150, opened 2025-04-19
fixing invalid urls for nvidia

chasebolt:
The doc examples are using _ but the release files are using -. Updating the docs to use - as that seems to be more consistent throughout.


t-lo (approved):
LGTM, thank you!

t-lo:
Good catch! Thank you Chase.

PR #152, opened 2025-04-22
Gasket driver

fl0eb:
# New Gasket driver management extension

* This SysExt provides a system extension for managing the Google Gasket driver on Flatcar.
* It is build similarly to the bundled Nvidia driver scripts.
* Similarly it uses a developer container to build the gasket and apex kernel extensions which are then installed loaded on the host
* Please be aware that Google more or less abandoned the project
* Therefor I decided to use https://github.com/KyleGospo/gasket-dkms as source since it includes multiple fixed for newer kernels

## How to use

To test this extension a Coral TPU M.2 Accelerator is required.
After the service has installed the driver, the device /dev/apex_0 should appear

## Testing done

* Manually deployed the extension into /opt/extensions
* Registered the extension with `ln -s /opt/extensions/gasket/gasket.raw /etc/extensions/gasket.raw`
* Registered the service with `sudo systemctl enable gasket`
* Using the device through hostPath mounts in containers (e.g. frigate)

This was done on flatcar instances with and without the hardware installed

**Todo**:


t-lo:
@fl0eb Thank you for your contribution!

I'm not sure the bakery is the right place for a kernel driver - extensions hosted here should not depend on a specific Flatcar version, and are ideally re-usable on distributions other than Flatcar. We don't have a concept of updating Bakery extensions in lock-step with the base OS so this can lead to breakage long after provisioning.

That said, the driver sure would be a great addition to the project. Flatcar does support "system-dependent" sysexts that are built and published with each release. Furthermore, there is safeguarding logic in our update client that will make sure to fetch updated extension images when a new base OS update is released, so "system-dependent" extensions continue working with new OS releases.

System-dependent extensions differ significantly from Bakery extensions, and are a bit more challenging. These extensions are part of Flatcar's core distribution repository. If you want to tackle this we would gladly accept your contribution! Steps involve:
- Get accustomed to the Flatcar SDK as we'll be doing core distro work: https://www.flatcar.org/docs/latest/reference/developer-guides/sdk-modifying-flatcar/
- Add an ebuild for the gasket kmod to the scripts repo: https://github.com/flatcar/scripts/tree/main/sdk_container/src/third_party/coreos-overlay
  - the ebuild should have a build-time dependency on configured kernel sources (package `coreos-modules`) for an off-tree build
- Successfully build the kmod with the SDK
- Add the kmod ebuild to the list of sysexts to be built at OS image build time: https://github.com/flatcar/scripts/blob/main/build_library/extra_sysexts.sh#L12
- Optionally, add a test (can be basic, like insmod'ing the kmod): https://github.com/flatcar/mantle/tree/main/kola

Sadly, the above is scarcely documented at this time, please chat us up on https://app.element.io/#/room/#flatcar:matrix.org for help. Would be awesome if you could take on this task.

Leaving the PR open for the time being so you have a place to respond.

fl0eb:
Hi @t-lo , thanks for your input! I do see the value in the system-dependent sysext, but I'm a bit concerned about potential build issues. The gasket driver is effectively abandonware and mainly kept alive by the community (e.g. [Issue](https://github.com/google/gasket-driver/issues/36) / [Fix](https://github.com/KyleGospo/gasket-dkms/pull/14) ) Integrating it into the main release pipeline might cause more trouble than it solves. That said, I’m open to supporting this direction. I’ll likely be able to start working on it in about two weeks. I'm closing this PR and I’ll open a new one with the updated approach once I can focus on it.

PR #153, opened 2025-04-23
nvidia-runtime: create symlinks dynamically

danzatt:
The sysext contains /opt/bin/nvidia-smi -> /usr/bin/nvidia-smi symlink. This does not play well with the prebuilt sysext, as it already ships the binaries in /usr/bin. Create the symlink dynamically, only when the prebuilt sysext is not loaded.

### DIFF
--- a/nvidia-runtime.sysext/create.sh
+++ b/nvidia-runtime.sysext/create.sh
@@ -53,6 +53,5 @@ function populate_sysext_root() {
   cp -aR out/usr/lib/*-linux-gnu/* "${sysextroot}/usr/lib/"
 
   ln -s /opt/nvidia "${sysextroot}/usr/local/nvidia"
-  ln -s /opt/bin/nvidia-smi "${sysextroot}/usr/bin/nvidia-smi"
 }
 # --
--- a/nvidia-runtime.sysext/files/usr/lib/systemd/system/nvidia.service.d/10-persistenced.conf
+++ b/nvidia-runtime.sysext/files/usr/lib/systemd/system/nvidia.service.d/10-persistenced.conf
@@ -1,8 +1,4 @@
 [Service]
-ExecStartPre=-/bin/sh -c "rm /run/extensions/nvidia-driver && systemctl restart systemd-sysext"
-ExecStartPost=-/opt/bin/nvidia-persistenced
-ExecStartPost=-/bin/sh -c "chcon -R -t container_file_t /dev/nvidia*"
-ExecStartPost=mkdir -p /run/extensions
-ExecStartPost=ln -sf /opt/nvidia/current /run/extensions/nvidia-driver
-ExecStartPost=systemctl restart systemd-sysext
+ExecStartPre=/usr/local/bin/_nvidia-runtime-helper pre
+ExecStartPost=/usr/local/bin/_nvidia-runtime-helper post
 ExecStopPost=rmmod nvidia_uvm nvidia_modeset nvidia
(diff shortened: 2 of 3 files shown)


nvidia-runtime.sysext/files/usr/local/bin/_nvidia-persistenced-helper
```diff
@@ -0,0 +1,47 @@
...
+  if [ "$SYSEXT_MODE" -eq 0 ]; then
+    rm /run/extensions/nvidia-driver && systemctl restart systemd-sysext || true
+  fi
+}
+
+_post() {
+  chcon -R -t container_file_t /dev/nvidia* || true
+
+  if [ "$SYSEXT_MODE" -eq 0 ]; then
+    /opt/bin/nvidia-persistenced || true
```
jepio, line 19: i've been meaning to drop launching `nvidia-persistenced` - it causes issues for some usecases. Let's leave this to users.

nvidia-runtime.sysext/files/usr/local/bin/_nvidia-persistenced-helper
```diff
@@ -0,0 +1,47 @@
...
+    mkdir -p /run/extensions
+    # /opt/nvidia/current gets mounted as an overlay on root filesystem
+    mkdir -p /opt/nvidia/current/usr/bin && ln -sf /opt/bin/nvidia-smi /opt/nvidia/current/usr/bin/nvidia-smi
+    mkdir -p /opt/nvidia/current/usr/local/ && ln -sf /opt/nvidia /opt/nvidia/current/usr/local/nvidia
+    ln -sf /opt/nvidia/current /run/extensions/nvidia-driver
+    systemctl restart systemd-sysext
+  else
+    systemctl enable --now nvidia-persistenced
```
jepio, line 27: i've been meaning to drop launching `nvidia-persistenced` - it causes issues for some usecases. Lets drop this line and leave this to users.

nvidia-runtime.sysext/files/usr/local/bin/_nvidia-persistenced-helper
```diff
@@ -0,0 +1,47 @@
...
+
+_post() {
+  chcon -R -t container_file_t /dev/nvidia* || true
+
+  if [ "$SYSEXT_MODE" -eq 0 ]; then
+    /opt/bin/nvidia-persistenced || true
+    mkdir -p /run/extensions
+    # /opt/nvidia/current gets mounted as an overlay on root filesystem
+    mkdir -p /opt/nvidia/current/usr/bin && ln -sf /opt/bin/nvidia-smi /opt/nvidia/current/usr/bin/nvidia-smi
+    mkdir -p /opt/nvidia/current/usr/local/ && ln -sf /opt/nvidia /opt/nvidia/current/usr/local/nvidia
```
jepio, line 23: were you planning on moving these lines to the `nvidia-driver.service`? Since it's only in that case that we mount `/opt/nvidia/current` as a sysext.

Second question: does sysext_mode ship `/usr/local/nvidia`?

jepio (approved):
lgtm but drop the persistenced and i have that one question around the symlinks

nvidia-runtime.sysext/files/usr/local/bin/_nvidia-persistenced-helper
jepio: `/usr/local/nvidia` appears to be used when we request that the GPU -Operator install the container-toolkit. So something (maybe symlinking `/usr/local/nvidia` to `/var/lib/nvidia`?) would be good to have in Flatcar

danzatt: Ok, returning the `/usr/local/nvidia` symlink to this sysext for now. We should investigate a more systematic approch.

PR #155, opened 2025-05-12
Add opkssh sysext

markafarrell:
# Add opkssh sysext

opkssh is a tool which enables ssh to be used with OpenID Connect allowing SSH access to be managed via identities like alice@example.com instead of long-lived SSH keys. It does not replace SSH, but instead generates SSH public keys containing PK Tokens and configures sshd to verify them. These PK Tokens contain standard [OpenID Connect ID Tokens](https://openid.net/specs/openid-connect-core-1_0.html). This protocol builds on the [OpenPubkey](https://github.com/openpubkey/openpubkey/blob/main/README.md) which adds user public keys to OpenID Connect without breaking compatibility with existing OpenID Provider.

https://github.com/openpubkey/opkssh

## How to use

1. Add sysext as per doc  (ensuring you change `my.email@gmail.com` to your actual gmail address
2. Run `opkssh login google` to generate private key and certificate
3. Login to host `ssh -o "IdentitiesOnly=yes" -i ~/.ssh/id_ecdsa.pub -i ~/.ssh/id_ecdsa core@my.flatcar.host`

## Testing done

Tested on Raspberry Pi 4 using instructions above

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -71,6 +71,7 @@ Check out documentation on specific extensions at the navigation menu on the lef
 | `nerdctl`        |  released    | [nerdctl versions](https://github.com/flatcar/sysext-bakery/releases/tag/nerdctl) |
 | `nvidia-runtime` |  released    | [nvidia-runtime versions](https://github.com/flatcar/sysext-bakery/releases/tag/nvidia-runtime) |
 | `ollama`         |  released    | [ollama versions](https://github.com/flatcar/sysext-bakery/releases/tag/ollama) |
+| `opkssh`         |  released    | [opkssh versions](https://github.com/flatcar/sysext-bakery/releases/tag/opkssh) |
 | `rke2`           |  released    | [rke2 versions](https://github.com/flatcar/sysext-bakery/releases/tag/rke2) |
 | `tailscale`      |  released    | [tailscale versions](https://github.com/flatcar/sysext-bakery/releases/tag/tailscale) |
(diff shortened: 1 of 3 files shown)


t-lo:
Thank you for your contribution @markafarrell , I tested the build locally and it works as expected.

Can we please add documentation for the new extension to `docs/opkssh.md` , and reference it in the table of available extensions in https://github.com/markafarrell/sysext-bakery/blob/feature/opkssh/docs/index.md#what-extensions-are-available ?

It might be worth going further (maybe in this PR or maybe in a follow-up PR) and automate the installation e.g. via a systemd unit (update `/etc/ssh/sshd_config` to use the module, and maybe other stuff from https://github.com/openpubkey/opkssh/blob/main/scripts/install-linux.sh). That way the extension could be installed and activated fully automatically.

markafarrell:
> Please add documentation in `docs/` as per my comment.

@t-lo sorry I forgot to push the docs.

Fixed now

markafarrell:
> It might be worth going further (maybe in this PR or maybe in a follow-up PR) and automate the installation e.g. via a systemd unit (update `/etc/ssh/sshd_config` to use the module, and maybe other stuff from https://github.com/openpubkey/opkssh/blob/main/scripts/install-linux.sh). That way the extension could be installed and activated fully automatically.

I'm not sure its really worth the effort. The butane snippet to get it configured and working is pretty small

markafarrell:
@t-lo the releases don't seem to be getting created for this. Are you able to have a look?

PR #159, opened 2025-06-03
Bump cilium cli

reavessm:
# Add new releases for cilium cli

Add new releases for cilium cli

## How to use

Make sure versions build

## Testing done

`./bakery.sh create cilium v0.18.4 --sysupdate true`


t-lo:
I'll close this PR as there is no need to explicitly pin these cilium versions (both v0.18.3 and v0.18.4 are released now).

PR #162, opened 2025-06-20
docs/kubernetes: fix component name

tormath1:
It was initially done here: https://github.com/flatcar/sysext-bakery/blob/9ebba24d11e66145f1c4f14c0557ac5aa4a3ea75/docs/kubernetes.md#usage

I think it's a side effect of the sysext-bakery refactoring.


chewi (approved):
I'm glad you figured that out because Thilo and I were really confused! :sweat_smile:

PR #164, opened 2025-06-23
Docker buildx

darkspadez:
# Add docker-buildx sysext

This adds the docker-buildx sysext to the repo. It is usually heavily used when running docker build/docker compose build. Currently docker compose build throws a warning that this is missing and this will fix that issue.


t-lo (approved):
Thank you for your contribution! Tested this with https://github.com/flatcar/sysext-bakery/pull/166, works well.

PR #175, opened 2025-08-01
Add support for generating signed sysexts

danzatt:
# Sign published sysexts

Use systemd-repart for creating signed DDI sysexts. The keys are expected to be stored in GitHub secrets SYSEXT_PRIVATE_KEY and SYSEXT_CERTIFICATE.

Note/TODO: this needs additional changes in the GH repo setup, there needs to be a signing key and certificate in GH secrets. Additionally, the signing certificate should be published as a release, so that consumers can use it to verify the sysexts. This can be automated through new GH actions, so that the signing key is would be generated and uploaded right in GH actions.

## How to use

The following butane config can be used for testing (pulls wasmedge sysext as an example, puts signing certificate into `/etc/verity.d` and forces sysext signature verification using service dropin). **For the sysext verification to work, you need to recompile kernel with `CONFIG_DM_VERITY_VERIFY_ROOTHASH_SIG` (implemented in a separate PR https://github.com/flatcar/scripts/pull/3162).**

Note: it uses releases from [my fork of sysext-bakery](github.com/danzatt/sysext-bakery/)

```
variant: flatcar
version: 1.0.0

storage:
  files:
    - path: /opt/extensions/wasmedge-0.14.1-x86-64.raw
      mode: 0420
      contents:
        source: https://github.com/danzatt/sysext-bakery/releases/download/wasmedge-0.14.1/wasmedge-0.14.1-x86-64.raw
    - path: /etc/sysupdate.wasmedge.d/wasmedge.conf
      contents:
        source: https://extensions.flatcar.org/extensions/wasmedge.conf
    - path: /etc/verity.d/sysext-bakery.crt
      mode: 0420
      contents:
        source: https://github.com/danzatt/sysext-bakery/releases/download/signing-key/sysext.crt
  links:
    - target: /opt/extensions/wasmedge-0.14.1-x86-64.raw
      path: /etc/extensions/wasmedge.raw
      hard: false
systemd:
  units:
    - name: systemd-sysext.service
      enabled: true
      dropins:
        - name: force-signature.conf
          contents: |
            [Service]
            TimeoutStartSec=10
            ExecStart=
            ExecStart=systemd-sysext refresh --image-policy="root=verity+signed+absent:usr=verity+signed+absent"
            ExecReload=
            ExecReload=systemd-sysext refresh --image-policy="root=verity+signed+absent:usr=verity+signed+absent"
            [Unit]
            JobRunningTimeoutSec=5
```

## Testing done

### DIFF
--- a/flatcar-container-wrap.sh
+++ b/flatcar-container-wrap.sh
@@ -15,4 +15,4 @@ for VARNAME in $(bash -c "compgen -v | grep -vE \"$EXCLUDED_VARS\""); do
 done
 
 docker run "${ARGS[@]}" "${FLATCAR_SDK_CONTAINER}:$LATEST_FLATCAR_SDK_VERSION" \
-  bash -c "$*"
+  bash -c "sudo ./setup_scripts_repo.sh && $*"
--- a/setup_scripts_repo.sh
+++ b/setup_scripts_repo.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# clone scripts with a custom branch and set up symlinks,
+# where the run_sdk_container normally mounts stuff
+git -C /tmp clone --depth=1 https://github.com/flatcar/scripts.git -b danzatt/sign-sysexts-nopatch
+sudo rm -rf /mnt/host/source
+sudo mkdir -p /tmp/scripts/__build__/images
+sudo ln -s /tmp/scripts/sdk_container /mnt/host/source
+sudo ln -s /tmp/scripts/__build__/images /mnt/host/source/src/build
+sudo ln -s /tmp/scripts/ /mnt/host/source/src/scripts
+
+# reinstall packages changed in the branch
+sudo emerge systemd azure-keyvault-pkcs11
+sudo touch /etc/pkcs11/pkcs11.conf
+echo "module: /usr/lib64/pkcs11/azure-keyvault-pkcs11.so" | sudo tee /etc/pkcs11/modules/azure-keyvault-pkcs11.module


lib/generate.sh
```diff
@@ -99,6 +99,32 @@ function _create_sysupdate() {
...
+function _generate_signed_sysext() {
+  BASEDIR="$1"
+  OUTPUT_IMAGE="$2"
+  FS_FORMAT="$3"
+
+  # Create temporary working directory
+  WORKDIR=$(mktemp -d)
+  trap 'rm -rf "$WORKDIR"' RETURN
+
+  cp -r "/usr/lib/systemd/repart/definitions/sysext.repart.d/" "$WORKDIR"
```
pothos, line 111: Maybe we can keep a know-good copy of these files in a subfolder instead of depending on the host contents

lib/generate.sh
```diff
@@ -107,23 +133,15 @@ function _generate_sysext() {
...
-      mkfs."${format}" -E root_owner=0:0 -d "${basedir}" "${fname}"
-      resize2fs -M "${fname}"
-      ;;
-    squashfs)
-      mksquashfs "${basedir}" "${fname}" -all-root -noappend -xattrs-exclude '^btrfs.'
-      ;;
-    erofs)
-      mkfs.erofs "${fname}" "${basedir}"
-      ;;
+  (
```
pothos, line 136: I think we should make signing opt-in, so wrap the the direct invocations in one if branch?

pothos: This could also reuse the exported repart flag options env var.
FYI: I also noticed problems with erofs and have filed a PR to fix this and if merged earlier you might need to adapt the settings when rebasing here https://github.com/flatcar/sysext-bakery/pull/176

.github/workflows/release.yaml
```diff
@@ -86,6 +86,8 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           pushd bakery
+          echo "${{ secrets.SYSEXT_PRIVATE_KEY }}" > sysext.key
+          echo "${{ secrets.SYSEXT_CERTIFICATE }}" > sysext.crt
```
pothos, line 90: I think we should publish the cert as release artifact - and maybe include it in the sysupdate definition?

pothos: Sysupdate has an option to verify with GPG (which is recommended) but anyway, this then would have to be part of the Ignition config. I think for now we can also provision the cert via Ignition. If the cert doesn't change it sounds ok to have it in the OS but if it's expiring every now and then, having it in the OS might be more complex for managing the lifecycle?

lib/generate.sh
```diff
@@ -122,8 +140,52 @@ function _generate_sysext() {
...
+  if [[ ${signed_ddi} == true ]] ; then
+    extname="${extname}-signed-ddi"
+  fi
+
+  announce "Creating extension image '${fname}' and generating SHA256SUM"
+  (
+    export SYSTEMD_REPART_MKFS_OPTIONS_BTRFS="--mixed -m single -d single --shrink"
+    export SYSTEMD_REPART_MKFS_OPTIONS_EXT2="-E root_owner=0:0"
+    export SYSTEMD_REPART_MKFS_OPTIONS_EXT4="$SYSTEMD_REPART_MKFS_OPTIONS_EXT2"
+    export SYSTEMD_REPART_MKFS_OPTIONS_SQUASHFS="-all-root -noappend -xattrs-exclude '^btrfs.'"
```
pothos, line 164: This needs EROFS, too, and the BTRFS compression options, available when rebasing this branch

lib/generate.sh
```diff
@@ -122,8 +140,52 @@ function _generate_sysext() {
...
+
+    if [[ ${signed_ddi} == true ]] ; then
+      _generate_signed_sysext "${basedir}" "${fname}" "$format"
+    else
+      options_varname="SYSTEMD_REPART_MKFS_OPTIONS_${format^^}"
```
pothos, line 169: Can we call `_generate_unsigned_sysext` here instead of duplicating the mkfs invocations?

lib/generate.sh
```diff
@@ -110,33 +147,58 @@ function _generate_sysext() {
...
-      ;;
 
-  esac
+  VERSION=$({ mksquashfs -version || true ; } | head -n1 | cut -d " " -f 3)
+  SQUASHFS_ARG=(-all-root -noappend)
+  if _version_ge "${VERSION}" "4.6.1"; then
+    SQUASHFS_ARG+=('-xattrs-exclude' "'^btrfs.'")
+  fi
+
+  # Note: We didn't chown to root:root, meaning that the file ownership is left as is
```
pothos, line 164:
```suggestion
  # Note: We didn't chown to root:root, meaning that the file ownership is left as is with btrfs
```

pothos:
> * I had to pull off some tricks to make it work, because bakery itself runs containers (i.e. Docker-in-Docker)

Maybe unpriv. Podman could help here because it can be nested many times (with the right invocation options).

lib/generate.sh
```diff
@@ -110,33 +147,58 @@ function _generate_sysext() {
...
+  SYSTEMD_REPART_MKFS_OPTIONS_BTRFS="--mixed -m single -d single --shrink --compress zstd"
+  SYSTEMD_REPART_MKFS_OPTIONS_EXT2="-E root_owner=0:0"
+  SYSTEMD_REPART_MKFS_OPTIONS_EXT4="$SYSTEMD_REPART_MKFS_OPTIONS_EXT2"
+  SYSTEMD_REPART_MKFS_OPTIONS_SQUASHFS="${SQUASHFS_ARG[*]}"
+  # Compression recommendations from https://erofs.docs.kernel.org/en/latest/mkfs.html
+  # and forcing a zero UUID for reproducibility (maybe we could also hash the name and version)
+  SYSTEMD_REPART_MKFS_OPTIONS_EROFS="-zlz4hc,12 -C65536 -Efragments,ztailpacking -Uclear --all-root"
```
pothos, line 171:
```suggestion
  SYSTEMD_REPART_MKFS_OPTIONS_EROFS="-zlz4hc,12 -C65536 -Efragments,ztailpacking -U00000000-0000-0000-0000-000000000000 --all-root"
```
From #181

danzatt:
> Are you sure we actually need Azure CLI? This is normally only needed when signing manually. In CI, you're more likely to use a managed identity or a token set in an environment variable. We currently used a managed identity for signing the boot binaries in CI. You may see errors about the `az` command not being found, but that's only because it tries that before other mechanisms.

We need it because the job isn't running on an Azure VM with attached managed identity (like in our CI). As the job is running in GitHub it's not possible to directly assign an Azure identity to the job/worker. Instead I configure the managed identity to trust this github repo to provide federated identity to Azure (I basically followed this guide https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure-openid-connect). Azure then gives the github action a one-time token to login into the managed identity. The PKCS11 token than uses az to authenticate (using the `AzureCliCredential`).

chewi:
I figured managed identities wouldn't be possible, but I thought environment variables might be. They are, but service principal secrets are not recommended. However! I chatted with Copilot, and it has told me that if azure-keyvault-pkcs11 adds support for workload identities (one extra line), then it should work without Azure CLI. The only clunky part is you have to write the token out to a file first. It admits this part could be better, but it's still preferable over installing Azure CLI.

```c++
Azure::Identity::ChainedTokenCredential::Sources{
    std::make_shared<Azure::Identity::WorkloadIdentityCredential>(),
    std::make_shared<Azure::Identity::EnvironmentCredential>(),
    getClientSecretCredential(),
    std::make_shared<Azure::Identity::AzureCliCredential>(),
    std::make_shared<Azure::Identity::ManagedIdentityCredential>()});
```

```yaml
permissions:
  id-token: write
  contents: read

jobs:
  authenticate:
    runs-on: ubuntu-latest
    steps:
      - name: Request OIDC token and save to file
        id: oidc-token
        uses: actions/github-script@v7
        with:
          script: |
            const fs = require('fs');
            const token = await github.actions.getIDToken();
            fs.writeFileSync('/tmp/azure-oidc-token', token);
            return '/tmp/azure-oidc-token';
      - name: Set AZURE_FEDERATED_TOKEN_FILE
        run: echo "AZURE_FEDERATED_TOKEN_FILE=${{ steps.oidc-token.outputs.result }}" >> $GITHUB_ENV
```

I'm also wondering whether you really need a Ubuntu image for this. The Flatcar SDK already has everything needed... except Docker? I'm confused about why that is needed though.

chewi:
> I've just reworked the PR. Now instead of building the sysext directly using `systemd-repart`, I keep the original workflow and then wrap the `.raw` file into a signed DDI using custom config for `systemd-repart`. This final signing/wrapping step is done in latest Flatcar SDK container.

Sounds good, but why did you have to do that?

sign_sysext_keyvault.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
pothos:
```suggestion
set -euo pipefail
```
We should stop script execution on errors

lib/setup_azure_keyvault.sh
```diff
@@ -0,0 +1,166 @@
...
+  az role assignment create \
+    --assignee-object-id "$USER_OID" \
+    --assignee-principal-type User \
+    --role "$role" \
+    --scope "$RG_ID" \
+    -o none
+done
+
+echo "Waiting 30 seconds for the permissions to apply. If the script fails, please wait and run it again."
+sleep 30
```
pothos: Do we have any chance to poll for the completion instead of defining a wait time that may be exceeded?

sign_sysext_keyvault.sh
```diff
@@ -0,0 +1,15 @@
...
+SYSEXT_NAME="$1"
+KEYVAULT_CERT_NAME="$2"
+
+if [ -n "${AZURE_FEDERATED_TOKEN_FILE+x}" ]; then
+  echo "Obtaining OIDC token..."
+  token=$(curl -sSL \
+    -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
+    "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=api://AzureADTokenExchange" \
+    | jq -r '.value')
+  echo "$token" > "$AZURE_FEDERATED_TOKEN_FILE"
```
chewi:
```suggestion
  curl -sSL \
    -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" \
    "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=api://AzureADTokenExchange" \
    | jq -r '.value' > "$AZURE_FEDERATED_TOKEN_FILE"
```

PR #176, opened 2025-08-04
lib/generate.sh: Fix EROFS file UIDs and use compression

pothos:
The EROFS mkfs invocation missed the --all-root flag to make sure that the files are owned by root regardless of the current UID during image generation. By default EROFS also doesn't compress and we can use LZ4 as a lightweight compresson option (LZMA would also work). Add the --all-root flag and the compression flag to mkfs.erofs to use UID 0 and compress the contents.

## How to use

```
./bakery.sh create --format erofs wasmtime v35.0.0
```
And then copy via ssh to a Flatcar VM, move the file under `/etc/extensions/` and reload the sysexts.

## Testing done

The wasmtime binary is now owned by root and the resulting image is half the size thanks to compression.

### DIFF
--- a/wasmcloud.sysext/create.sh
+++ b/wasmcloud.sysext/create.sh
@@ -33,6 +33,7 @@ function populate_sysext_root() {
 
   mkdir -p "${sysextroot}/usr/bin"
   cp -a "wasmcloud-${rel_arch}-unknown-linux-musl" "${sysextroot}/usr/bin/wasmcloud"
+  chmod +x "${sysextroot}/usr/bin/wasmcloud"
   cp -a "nats-server-${nats}-linux-${go_arch}/nats-server" "${sysextroot}/usr/bin/"
 }
 # --


pothos:
Also added a patch to enable btrfs compression and add a note that something like `--all-root` does not exist for mkfs.btrfs and the resulting files may have unwanted UIDs set.

hsiangkao:
I suggest using `-zlz4hc,12` instead of `lz4` unless the build time is sensitive.

pothos:
> I suggest using `-zlz4hc,12` instead of `lz4` unless the build time is sensitive.

Thank you for the comment! Now I've checked the https://github.com/erofs/erofs-utils README and I see there, too. I also thought about LZMA because a 5.16 kernel could be a low enough requirement. What would be better in that case, lz4hc or LZMA? Also considering long term plans?

hsiangkao:
> Thanks, updated the mkfs.erofs options with the recommendations from the guide.
> 
> I've also noticed that by default it's not reproducible, and I had to add `-Uclear` to clear the filesystem UUID (maybe we could also calculate a hash based on the name and the version).

`-Uclear -T0 --mkfs-time` will make images  reproducible. @pothos

pothos:
By the way, while testing the compression flags in `mkfs.erofs` I had the idea that all mkfs tools (other filesystems have the same problems in this regard) could print what minimal kernel version (or kernel config option) the resulting image is compatible with. Maybe it's complicated to implement, but it would help against accidentally raising the kernel requirements when tweaking the options.

pothos:
@flatcar/flatcar-integrations Can you review the changes?

danzatt:
Looks good. The only slight worry I had was if systemd won't get confused if there are multiple sysexts with the same (zero) UUID. I've tested this on latest alpha and it works as expected, so I guess that's not an issue.

pothos:
Thanks, yes, I should have written that I've also tested this and that squashfs doesn't have one either and a zero UUID won't show up in `/dev/disk/by-uuid/`.

PR #179, opened 2025-08-22
Extensions: Add ig

burak-ok:
# Adds the IG sysext extension

This PR adds an extension for [Inspektor Gadget](https://inspektor-gadget.io/). Specifically the `ig` and `gadgetctl` binary. One can use `ig` to directly inspect/observe the linux host with all its containers.  Another usage might be to spawn an `ig daemon`, which then runs in the background, and control that daemon instance with `gadgetctl`

## How to use

```
$ ./bakery.sh create ig v0.43.0
$ ./bakery.sh boot ig.raw
core@localhost $ while true; do sleep 1 && date > /dev/null; done &
core@localhost $ sudo ig daemon &
core@localhost $ sudo gadgetctl run trace_exec --host
RUNTIME.CONTAINERNAME   COMM                 PID        TID TTY ARGS         ER…
                        date                2007       2007 0   /usr/bin/da…    
                        sleep               2008       2008 0   /usr/bin/sl…    
                        date                2009       2009 0   /usr/bin/da…    
                        sleep               2010       2010 0   /usr/bin/sl…    

```

## Testing done

See `How to use`

I copy pasted the butane snippet mostly from other doc files. How can I test it when the contents point to config files on webservers, which are not there (yet).

### DIFF
--- a/docs/ig.md
+++ b/docs/ig.md
@@ -1,0 +1,54 @@
+# Inspektor Gadget sysext
+
+This sysext ships [Inspektor Gadget](https://www.inspektor-gadget.io/), a collection of tools and framework for data collection and system inspection on Kubernetes clusters and Linux hosts using eBPF.
+
+This sysext includes both the `ig` and `gadgetctl` binaries:
+- `ig` - The main inspektor-gadget tool for running gadgets on Linux hosts
+- `gadgetctl` - Tool for managing gadgets and interacting with `ig` running in daemon mode
+
+## Usage
+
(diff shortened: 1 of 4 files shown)


burak-ok:
> I might use this sysext extension to replace the outdated "local-gadget" we are using in Flatcar test suite: https://github.com/flatcar/mantle/blob/main/kola/tests/bpf/local-gadget.go

Didn't even knew the old `local-gadget` existed there :D If there are any issues with `ig` feel free to ping me :)

PR #183, opened 2025-09-18
Fix docker-buildx binary name

pantherman594:
# Fix docker-buildx binary name

docker-buildx was being saved as docker-compose, which is overwritten when the docker compose sysext is also used

## How to use

## Testing done


t-lo (approved):
Did a test build with 0.28.0 and 0.10.0, used `bakery.sh boot` to verify. Works as advertised.

PR #184, opened 2025-09-26
wasmedge: fix build for version 0.15.1 and higher

t-lo:
WasmEdge tarball 0.15.0 and higher do not contain a WasmEdge.../ subdirectory anymore. Instead, the FS root is shipped directly.

This change fixes the wasmedge build for versions >=0.15.0 and adds helper functions to compare semvers.

Tested with local builds, both wasmedge-0.14.1 and wasmedge-0.15.0.

### DIFF
--- a/wasmedge.sysext/create.sh
+++ b/wasmedge.sysext/create.sh
@@ -26,7 +26,12 @@ function populate_sysext_root() {
   mkdir -p "${sysextroot}/usr/bin"
   mkdir -p "${sysextroot}/usr/lib"
 
-  cp -a "WasmEdge-${version}-Linux"/bin/wasmedge "${sysextroot}"/usr/bin/
-  cp -a "WasmEdge-${version}-Linux"/lib/* "${sysextroot}"/usr/lib/
+  local prefix="./"
+  if semver_lower "${version}" "0.15.0" ; then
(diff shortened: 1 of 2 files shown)


wasmedge.sysext/create.sh
```diff
@@ -26,7 +26,12 @@ function populate_sysext_root() {
   mkdir -p "${sysextroot}/usr/bin"
   mkdir -p "${sysextroot}/usr/lib"
 
-  cp -a "WasmEdge-${version}-Linux"/bin/wasmedge "${sysextroot}"/usr/bin/
-  cp -a "WasmEdge-${version}-Linux"/lib/* "${sysextroot}"/usr/lib/
+  local prefix="./"
+  if semver_lower "${version}" "0.15.0" ; then
+    local prefix="WasmEdge-${version}-Linux/"
+  fi
```
sayanchowdhury, line 32:
```suggestion
  local prefix="."
  if semver_lower "${version}" "0.15.0" ; then
    local prefix="WasmEdge-${version}-Linux"
  fi
```

Do we need to the trailing `/`?

t-lo: We do not, great catch!

sayanchowdhury (approved):

PR #191, opened 2025-11-22
Bump tailscale version to 1.90.8

gdonval:
# Bump tailscale version to 1.90.8

Bump is necessary to fix security issues

TS-2025-008: Tail Lock skip (in certain conditions).
TS-2025-007: race condition when using one-off auth keys.

## How to use

1.90.8 is the same version as latest for now. If latest works, this one will.
Otherwise install the sysext and see if node shows up on Tailscale website.

## Testing done

### DIFF
--- a/release_build_versions.txt
+++ b/release_build_versions.txt
@@ -71,7 +71,7 @@ ollama latest
 rke2 v1.32.2+rke2r1
 rke2 latest
 
-tailscale v1.76.6
+tailscale v1.90.8
 tailscale latest
 
 wasmcloud v1.0.0


t-lo:
@gdonval Please pardon the long wait, and thank you for your contribution!

I think we can just remove the pinned versions, 1.90.8 is available since it's been "latest" at some point, and `tailscale latest` will take care of always building newer versions when they are released.

t-lo:
Closing this as we regularly build new releases and the "old" tailscale is nothing more than a placeholder. Tailscale sysext releases include 1.90.8 proposed by this PR; the latest available sysext version from the Bakery is 1.98.3, which corresponds to the latest upstream release available (see https://github.com/flatcar/sysext-bakery/releases/tag/tailscale). @gdonval Please consider filing a PR that simply removes the outdated version entirely from `release_build_versions.txt`.

PR #204, opened 2026-02-05
Fix extremely minor type in bake_flatcar_image.sh

gitRaiku:
# Fix a small typo in bake_flatcar_image
Kubernetes was written wrong


tormath1:
Nice spot, thanks a lot for taking the time of fixing / submitting this. Let's hope it won't break any existing automation relying on this specific  example. 😂

PR #205, opened 2026-03-03
bakery boot: use alpha release by default

t-lo:
Alpha will expose bakery developers to the latest features while being fully tested (i.e. guaranteed ito work).


lib/boot.sh
tormath1:
```suggestion
  echo "The latest alpha OS image will be downloaded if no Flatcar image is present locally."
```

tormath1 (approved):

PR #208, opened 2026-03-27
release_build_versions: add opkssh

### LINKED ISSUE #206: opkssh is listed as an extension and isn't being built
opkssh builds are not happening - this was flagged in #155 but was missed as the issue was closed.  I suggest we either fix the builds or drop it from the published list.

tormath1:
without this the extension is not built by the automation

Closes: #206


danzatt (approved):
LGTM

PR #210, opened 2026-04-09
docs: report missing extension documentation drift

### LINKED ISSUE #207: Add CI workflow to detect extensions missing from documentation index
## Summary

When a new extension is added to `release_build_versions.txt` (i.e. it becomes part of the automated build and release pipeline), there is no automated check to ensure it also gets a corresponding entry in the documentation at `docs/index.md` and a dedicated `docs/<extension>.md` page. This can lead to extensions being built and released but not discoverable via https://flatcar.github.io/sysext-bakery/.

## Proposal

Add a GitHub Actions workflow that runs whenever `release_build_versions.txt` is pushed to `main`. The workflow would:

1. Parse the extension names from `release_build_versions.txt`.
2. Parse the extension names listed in the table in `docs/index.md`.
3. For each extension present in the build list but absent from the docs index, open a GitHub issue — **if** an open issue for that extension does not already exist.

### Issue deduplication

To avoid duplicate issues, each opened issue should use a consistent title pattern (e.g. `docs: add documentation for <extension> sysext`) and/or a label (e.g. `missing-docs`). Before opening, the workflow should search for existing open issues matching the extension name and that label.

### Issue content

Each opened issue should include:
- The extension name
- A note that it was detected as present in `release_build_versions.txt` but missing from `docs/index.md`
- A checklist of what's needed:
  - [ ] Add a row to the extensions table in `docs/index.md`
  - [ ] Create `docs/<extension>.md` with a description, upstream project link, and a Butane configuration snippet
  - [ ] Verify the extension's releases are accessible at `https://github.com/flatcar/sysext-bakery/releases/tag/<extension>`

### Scope

This workflow intentionally does **not** auto-generate documentation or open PRs with draft docs, since each extension's docs page requires human-written content (upstream project description and link, extension-specific configuration notes, and a tested Butane snippet).

The reverse case — an extension removed from `release_build_versions.txt` that still has docs listing it as "released" — is out of scope for now but could be added later.

## Existing gap

For reference, there are extensions today that appear in `release_build_versions.txt` but are not listed in the `docs/index.md` table (e.g. `btop`, `tilde`). This workflow would have caught those.

officialasishkumar:
# docs: report missing extension documentation drift

Add a workflow that watches `release_build_versions.txt` on `main` and opens deduplicated issues whenever an extension is part of the automated release pipeline but its documentation is incomplete.

The checker parses the release list, the docs index table, and the existing docs pages. It also accepts the repository's established underscore filenames for hyphenated extensions so existing pages like `docker_buildx.md`, `docker_compose.md`, and `nvidia_runtime.md` do not get reported as false positives.

Closes: #207

## How to use

Run the new `Report missing extension documentation` workflow manually, or merge a change to `release_build_versions.txt` on `main`. The workflow will inspect the release list and open one issue per undocumented extension unless an open issue with the generated title already exists.

For local verification, run:
`./tools/report_missing_extension_docs.sh`

## Testing done

Ran:
`bash -n tools/report_missing_extension_docs.sh`

Ran:
`./tools/report_missing_extension_docs.sh`

Output:
```
Missing documentation detected for 'btop'
  docs/index.md entry missing: true
  dedicated docs page missing: true
  Dry run only; no issue created.
Missing documentation detected for 'tilde'
  docs/index.md entry missing: true
  dedicated docs page missing: true
  Dry run only; no issue created.
```

The changelog and image-diff checklist items below are not applicable for this workflow-only change.

### DIFF
--- a/tools/report_missing_extension_docs.sh
+++ b/tools/report_missing_extension_docs.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# Detect extensions that are part of the automated release pipeline but are
+# missing from the documentation index and/or do not have a dedicated docs page.
+
+set -euo pipefail
+
+scriptroot="$(cd "$(dirname "${BASH_SOURCE[0]}")/.."; pwd)"
+source "${scriptroot}/lib/helpers.sh"
+
+create_issues="$(get_optional_param "create-issues" "false" "${@}")"
+repository="${GITHUB_REPOSITORY:-flatcar/sysext-bakery}"
+
+function list_release_extensions() {
+  sed -e 's:\s*#.*::' -e 's/[[:space:]]*$//' -e '/^$/d' \
+    "${scriptroot}/release_build_versions.txt" \
+  | awk '{print $1}' \
+  | sort -u
+}
+# --
+
+function list_index_extensions() {
+  sed -n -E 's/^\|[[:space:]]*`([^`]+)`[[:space:]]*\|.*$/\1/p' \
+    "${scriptroot}/docs/index.md" \
+  | sort -u
+}
+# --
+
+function docs_page_path() {
+  local extension="$1"
+  local candidate
+
+  # The repo usually uses docs/<extension>.md, but a few established pages map
+  # hyphens to underscores instead.
+  for candidate in "${extension}.md" "${extension//-/_}.md"; do
+    if [[ -f "${scriptroot}/docs/${candidate}" ]] ; then
+      echo "${candidate}"
+      return 0
+    fi
+  done
+
+  return 1
+}
+# --
+
+function issue_title() {
+  local extension="$1"
+  echo "docs: add documentation for ${extension} sysext"
+}
+# --
+
+function issue_body() {
+  local extension="$1"
+  local missing_index="$2"
+  local missing_page="$3"
+
+  cat <<EOF
+This extension was detected in \`release_build_versions.txt\`, but its documentation is incomplete.
+
+Current status:
+- Extension: \`${extension}\`
+- Missing from \`docs/index.md\`: \`${missing_index}\`
+- Missing dedicated docs page: \`${missing_page}\`
+
+Checklist:
+- [ ] Add a row to the extensions table in \`docs/index.md\`
+- [ ] Create \`docs/<extension>.md\` (or the repo's established underscore variant for hyphenated extensions) with a description, upstream project link, and a Butane configuration snippet
+- [ ] Verify the extension's releases are accessible at \`https://github.com/flatcar/sysext-bakery/releases/tag/${extension}\`
+EOF
+}
+# --
+
+declare -A index_extensions=()
+
+while IFS= read -r extension; do
+  index_extensions["${extension}"]="true"
+done < <(list_index_extensions)
+
+open_issues_json="$(gh issue list -R "${repository}" --state open --limit 200 --json title)"
+missing_extensions=()
+
+while IFS= read -r extension; do
+  missing_index="false"
+  missing_page="false"
+
+  if [[ -z ${index_extensions["${extension}"]+x} ]] ; then
+    missing_index="true"
+  fi
+
+  if ! docs_page_path "${extension}" >/dev/null ; then
+    missing_page="true"
+  fi
+
+  if [[ ${missing_index} == false && ${missing_page} == false ]] ; then
+    continue
+  fi
+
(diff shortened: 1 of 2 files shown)


t-lo:
@bexelbie Could you please have a look and check if this PR addresses your issue #207 ?

bexelbie:
I did not test the code and only reviewed it by reading it. This seems to solve the problem that I raised and I like the fix. 

Is this repository one that is regularly reviewed in one of the community meetings? I'm a lax community member and so I don't make enough of the meetings to know for sure. The reason I'm asking is because if we open issues that no one will ever see then we're just changing the shape of the problem instead of solving it.

t-lo:
> I did not test the code and only reviewed it by reading it. This seems to solve the problem that I raised and I like the fix.

I'll run a test later today to verify.

> Is this repository one that is regularly reviewed in one of the community meetings? I'm a lax community member and so I don't make enough of the meetings to know for sure. The reason I'm asking is because if we open issues that no one will ever see then we're just changing the shape of the problem instead of solving it.

We maintain a GitHub team - "Flatcar Integrations" and regularly check on issues and PRs here. Issues and PRs can also be added to the Tactical board (https://github.com/orgs/flatcar/projects/7/views/1) but do not land there automatically at this point.

Also, we could auto-label these docs issues with "good first issue" because they are :wink: .

tools/report_missing_extension_docs.sh
```diff
@@ -0,0 +1,124 @@
...
+}
+# --
+
+declare -A index_extensions=()
+
+while IFS= read -r extension; do
+  index_extensions["${extension}"]="true"
+done < <(list_index_extensions)
+
+open_issues_json="$(gh issue list -R "${repository}" --state open --limit 200 --json title)"
```
t-lo: Can we drop gh usage in the script and instead use a GH action to create issues?

E.g. https://github.com/marketplace/actions/issues-helper#find-issues can create issues if these do not exist yet. I think this script should just output a list of issues to be created and pass creation to a github action.

tools/report_missing_extension_docs.sh
```diff
@@ -0,0 +1,124 @@
...
+  fi
+
+  missing_extensions+=( "${extension}" )
+
+  echo "Missing documentation detected for '${extension}'"
+  echo "  docs/index.md entry missing: ${missing_index}"
+  echo "  dedicated docs page missing: ${missing_page}"
+
+  title="$(issue_title "${extension}")"
+  if jq -e --arg title "${title}" '.[] | select(.title == $title)' \
```
t-lo: Same as above - let's move this logic to a github action that already implements "create issue if it does not exists".

tools/report_missing_extension_docs.sh
```diff
@@ -0,0 +1,124 @@
...
+    echo "  Open issue already exists: ${title}"
+    continue
+  fi
+
+  if [[ ${create_issues} != true ]] ; then
+    echo "  Dry run only; no issue created."
+    continue
+  fi
+
+  gh issue create \
```
t-lo: Same - no need to call `gh` from the script; this will make the script more flexible and self contained, since no github interaction is required to run it.

t-lo:
@officialasishkumar Thank you for working on this! Looks good overall; I left a few comments re: `gh` usage in the script (I'd rather we don't). Could you please have a look at the https://github.com/marketplace/actions/issues-helper action which largely implements what we're trying to do with `gh` in the script?

Would love to get this in shape for merge.

t-lo (changes requested):
Looks good in general. main issue is with the script requiring `gh` to run (which makes it harder to use locally). Let's move that into a github action step that uses an action to post the issue.

tools/report_missing_extension_docs.sh
officialasishkumar: Moved issue creation into the workflow and left the script as JSON output only in 2b8711f.

tools/report_missing_extension_docs.sh
officialasishkumar: Moved issue creation into the workflow and left the script as JSON output only in 2b8711f.

tools/report_missing_extension_docs.sh
officialasishkumar: Moved issue creation into the workflow and left the script as JSON output only in 2b8711f.

t-lo (approved):
Thank you @officialasishkumar , LGTM!

PR #203, opened 2026-01-22
Add sched-ext/scx

ananthb:
# Run userspace schedulers with sched-ext/scx

Add the scx repository to run sched-ext/scx userspace schedulers.
sched-ext is a Kernel subsystem that allows running task schedulers
in userspace. This repository adds a host of userspace schedulers and
tooling to load, run, and unload them.

## How to use

**TODO**

## Testing done

[Describe the testing you have done before submitting this PR. Please include both the commands you issued as well as the output you got.]

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #212, opened 2026-05-30
Add sched-ext/scx

ananthb:
# Run userspace schedulers with sched-ext/scx

Add the `scx` sysext that ships the [sched-ext userspace schedulers](https://github.com/sched-ext/scx) plus the `scxtop` and `scxcash` tools. sched-ext is a Linux kernel subsystem that allows running task schedulers in userspace. The upstream repo provides a set of reference schedulers and tooling to load and observe them.

This re-opens #203. All of @t-lo's review feedback from that thread is applied here:

- Build inside an Alpine container; install `zlib-static` / `zstd-static` via apk and drop the from-source fallback for libz / libzstd.
- Use `CARGO_BUILD_TARGET=$(rustc -vV ... host)` for cross-compile consistency, and set the target-prefixed `CARGO_TARGET_<TRIPLE>_RUSTFLAGS` / `CARGO_TARGET_<TRIPLE>_PKG_CONFIG_ALL_STATIC` / `CARGO_TARGET_<TRIPLE>_LIBBPF_SYS_STATIC` env vars so the static flags actually apply when `CARGO_BUILD_TARGET` is set. Link statically with `-C target-feature=+crt-static` and `-C link-arg=-Wl,-static`.
- Install from `target/<triple>/release/` and widen the install glob from `scx_*` to `scx*` so `scxtop` and `scxcash` come along.
- Document the full scheduler set and the bundled tools in `docs/scx.md` and `files/usr/share/scx/scx`.
- Register `scx` in `docs/index.md` and `release_build_versions.txt` so the release Action picks it up.

## How to use

Drop the sysext via Ignition (snippet in `docs/scx.md`). By default the bundled `scx.service` runs `scx_bpfland`. Override by writing `/etc/default/scx`:

```
SCX_SCHEDULER_OVERRIDE=scx_lavd
SCX_FLAGS_OVERRIDE=-a
```

…then `systemctl restart scx.service`. The unit is gated by `ConditionPathIsDirectory=/sys/kernel/sched_ext`, so it stays inert on kernels without sched_ext.

## Testing done

This update applies the suggested build-script fixes verbatim from the prior review. Yet to do another round of local boot tests.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -81,6 +81,7 @@ Check out documentation on specific extensions at the navigation menu on the lef
 | `ollama`         |  released    | [ollama versions](https://github.com/flatcar/sysext-bakery/releases/tag/ollama) |
 | `opkssh`         |  released    | [opkssh versions](https://github.com/flatcar/sysext-bakery/releases/tag/opkssh) |
 | `rke2`           |  released    | [rke2 versions](https://github.com/flatcar/sysext-bakery/releases/tag/rke2) |
+| `scx`            |  released    | [scx versions](https://github.com/flatcar/sysext-bakery/releases/tag/scx) |
 | `tailscale`      |  released    | [tailscale versions](https://github.com/flatcar/sysext-bakery/releases/tag/tailscale) |
 | `wasmcloud`      |  released    | [wasmcloud versions](https://github.com/flatcar/sysext-bakery/releases/tag/wasmcloud) |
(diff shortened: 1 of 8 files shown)


ananthb:
I'll be back next weekend to test this out properly.

t-lo (approved):
Tested for both amd64 and arm64 (build and boot, run `scxtop`) - it's working!

t-lo:
Merging - thank you @ananthb for bringing this over the finish line!

ananthb:
Thank you for all of the support!

PR #226, opened 2026-06-29
coder, haproxy: add system extensions

ananthb:
## Summary

- Adds two new sysexts: `coder` and `haproxy`.
- **coder**: packages the upstream `coder_<v>_linux_<arch>.tar.gz` and ships `coder.service` / `coder-workspace-proxy.service` verbatim from the upstream Debian package. A `sysusers.d` entry creates the `coder` system user the units run as; `tmpfiles.d` seeds default env files into `/etc/coder.d/` on first boot. The tarball is verified against the per-release `coder_<v>_checksums.txt`.
- **haproxy**: builds a statically linked `haproxy` from upstream source inside Alpine with OpenSSL, PCRE2 (JIT), zlib, threads and the Prometheus exporter. Ships the upstream `admin/systemd/haproxy.service.in` with the documented sandboxing options enabled (`NoNewPrivileges`, `ProtectSystem=strict`, restricted capabilities/syscalls). `sysusers.d`/`tmpfiles.d` set up the `haproxy` user and a minimal default `/etc/haproxy/haproxy.cfg`.
- Wires both into `release_build_versions.txt` and the docs index, with usage docs at `docs/coder.md` and `docs/haproxy.md`.

## Test plan

- [ ] `./bakery.sh list coder` lists upstream `vX.Y.Z` releases
- [ ] `./bakery.sh list haproxy` lists upstream stable `X.Y.Z` releases (no `-devN`/`-rcN`)
- [ ] `./bakery.sh create coder vX.Y.Z` produces working `coder.raw` for x86-64 and arm64
- [ ] `./bakery.sh create haproxy X.Y.Z` produces working `haproxy.raw` for x86-64 and arm64 (static binary)
- [ ] `./bakery.sh boot coder` boots Flatcar, `coder --version` works, `coder.service` starts once `/etc/coder.d/coder.env` is populated
- [ ] `./bakery.sh boot haproxy` boots Flatcar, `haproxy -v` works, `haproxy.service` starts cleanly under the hardened sandbox with the seeded default config

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #79, opened 2024-07-01
add haproxy sysext creation script

justdan96:
# Add HAProxy Sysext Creation Script

This PR adds an HAProxy creation script to the repo. The creation has been tested and the running of the system extension tested on a VM. The script is heavily based on the keepalived script. Options and sample configurations have been pulled from AlmaLinux sources.

## How to use

Just run `./create_haproxy_sysext.sh` on a Linux system with Docker installed.

## Testing done

Script run and system extension installed on Ubuntu system. 

- [ ] Changelog entries added in the respective `changelog/` directory (user-facing change, bug fix, security fix, update)
- [ ] Inspected CI output for image differences: `/boot` and `/usr` size, packages, list files for any missing binaries, kernel modules, config files, kernel modules, etc.

PR #228, opened 2026-06-29
haproxy: add system extension

ananthb:
## Summary

Packages [HAProxy](https://www.haproxy.org/), a reliable high-performance TCP/HTTP load balancer.

- Builds `haproxy` from upstream source inside an ephemeral `alpine:3.21` container as a fully static musl binary with OpenSSL (incl. `USE_QUIC_OPENSSL_COMPAT=1` for HTTP/3), PCRE2 (JIT), Lua 5.4, zlib, threads, Linux capabilities, transparent proxy, network namespaces, TCP Fast Open, the Prometheus exporter and accept4/prctl enabled.
- No runtime library dependencies on the host: the binary is self-contained and drops into `/usr/bin/haproxy`. Follows the same static-alpine pattern as `bird.sysext` and `chrony.sysext` in this repo (originally prototyped in #79).
- Ships the upstream `admin/systemd/haproxy.service.in` with sandboxing turned on: `User=haproxy`, `NoNewPrivileges`, `ProtectSystem=strict`, capability/syscall filters. PID file and master socket live under `/run/haproxy/` (via `RuntimeDirectory=haproxy`). `Type=simple` — we don't link libsystemd so `sd_notify` isn't available.
- `sysusers.d` creates the `haproxy` system user; `tmpfiles.d` seeds a minimal default `/etc/haproxy/haproxy.cfg` (stats on `127.0.0.1:8404`) on first boot.
- Version listing walks the upstream `/download/<X.Y>/src/releases.json` index, filtering out `-devN` / `-rcN`.
- Wires `haproxy latest` into `release_build_versions.txt` and adds `docs/haproxy.md` plus an index row.

Split out of #226.

## Test plan

- [ ] `./bakery.sh list haproxy` lists upstream stable releases (no `-dev` / `-rc`)
- [ ] `./bakery.sh create haproxy X.Y.Z` produces a working `haproxy.raw` for x86-64 and arm64
- [ ] `./bakery.sh boot haproxy` boots Flatcar, `haproxy -v` shows OpenSSL/Lua/PCRE2 enabled, `ldd /usr/bin/haproxy` reports "not a dynamic executable", and `haproxy.service` starts cleanly under the hardened sandbox with the seeded default config
- [ ] Lua script in `/etc/haproxy/haproxy.cfg` loads (smoke test that Lua is compiled in)


haproxy.sysext/files/usr/lib/systemd/system/haproxy.service
```diff
@@ -0,0 +1,41 @@
...
+Description=HAProxy Load Balancer
+After=network-online.target
+Wants=network-online.target
+
+[Service]
+EnvironmentFile=-/etc/default/haproxy
+EnvironmentFile=-/etc/sysconfig/haproxy
+Environment="CONFIG=/etc/haproxy/haproxy.cfg" "PIDFILE=/run/haproxy/haproxy.pid" "EXTRAOPTS=-S /run/haproxy/haproxy-master.sock"
+ExecStart=/usr/bin/haproxy -Ws -f $CONFIG -p $PIDFILE $EXTRAOPTS
+ExecReload=/usr/bin/haproxy -Ws -f $CONFIG -c $EXTRAOPTS
```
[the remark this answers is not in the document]
ananthb: Tracking upstream `admin/systemd/haproxy.service.in` verbatim here — keeping the ExecReload as upstream ships it.

t-lo:
Thank you @ananthb , this is really useful. I'd much prefer if we could build `haproxy` statically so we do not have any library dependencies in the system extension. In general, shipping statically linked binaries whenever possible is preferred for sysexts.

There is prior work here: https://github.com/flatcar/sysext-bakery/pull/79/changes#diff-a31f2be94cde370b5bc113418a178ba6a4fd5875bef5105b94938d3ea4266b7b in an outdated PR that was filed before we added build wrappers `bakery.sh` and `lib`, but I think the approach there (building a static haproxy in a transient alpine container) works better. Do you think you could update your PR to do it that way?

t-lo (changes requested):
Let's convert this to a static alpine build just like the outdated PR (https://github.com/flatcar/sysext-bakery/pull/79) did it.

ananthb:
Switched to a fully static musl build in `alpine:3.21`.

haproxy.sysext/build.sh
```diff
@@ -0,0 +1,61 @@
...
+
+make -j"$(nproc)" \
+  TARGET=linux-musl \
+  USE_OPENSSL=1 \
+  USE_QUIC=1 USE_QUIC_OPENSSL_COMPAT=1 \
+  USE_PCRE2=1 USE_PCRE2_JIT=1 \
+  USE_ZLIB=1 \
+  USE_LUA=1 LUA_LIB_NAME=lua5.4 \
+    LUA_INC=/usr/include/lua5.4 \
+    LUA_LDFLAGS='-L/usr/lib/lua5.4 -l:liblua.a' \
```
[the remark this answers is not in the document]
ananthb: I checked a running alpine container and copilot is wrong here

haproxy.sysext/build.sh
```diff
@@ -0,0 +1,61 @@
...
+cd "${src}"
+
+curl -fsSLO "https://www.haproxy.org/download/${branch}/src/haproxy-${version}.tar.gz"
+curl -fsSLO "https://www.haproxy.org/download/${branch}/src/haproxy-${version}.tar.gz.sha256"
+sha256sum -c "haproxy-${version}.tar.gz.sha256"
+
+tar -xf "haproxy-${version}.tar.gz"
+cd "haproxy-${version}"
+
+make -j"$(nproc)" \
```
t-lo, line 39: The build will error out / print a massive amount of warnings for versions prior to 3.0.0, mostly because of outdated library headers / dependencies (most notably OpenSSL).

I think it's fine to limit support to builds starting from 3.0, but then we should not return older versions in `create.sh::list_available_versions`.

haproxy.sysext/create.sh
```diff
@@ -0,0 +1,71 @@
...
+  listing="$(curl -fsSL --retry-delay 1 --retry 60 --retry-connrefused \
+    --retry-max-time 60 --connect-timeout 20 \
+    "https://www.haproxy.org/download/")"
+
+  branches="$(echo "${listing}" \
+    | grep -oE 'href="[0-9]+\.[0-9]+/"' \
+    | sed -E 's|href="||;s|/"||' \
+    | sort -uV)"
+
+  for branch in ${branches} ; do
```
t-lo, line 35:
```suggestion
  for branch in ${branches} ; do
    if semver_lower "${branch}" "3.0" ; then
      continue
    fi
```
The `jq` command errs out for `1.0` since that version has an almost empty version JSON. Hence, `bakery.sh list haproxy` breaks. If we limit the output to newer versions this will be mitigated.

Also see the other comment on builds < 3.0 failing / printing lots of warnings from outdated lib dependencies - we can address this by not supporting builds of older versions, so we should filter everything before 3.0.

t-lo (changes requested):
This is getting into shape, thank you for putting in the work and converting it to a static Alpine build! Last remaining nit is on a build issue for older versions, please have a look.

haproxy.sysext/create.sh
ananthb: Done.

t-lo (approved):
Tested on x86-64 and ARM64, works great. Thank you!

PR #229, opened 2026-07-01
dataplaneapi: add system extension

ananthb:
## Summary

Packages the [HAProxy Data Plane API](https://www.haproxy.com/documentation/dataplaneapi/) — the REST API that manages an HAProxy instance's configuration and runtime state — as its own sysext. Intended to compose with the `haproxy` sysext in #228, but kept separate so upstream release cadences and per-host use cases stay decoupled.

- Installs the upstream `dataplaneapi` Go binary from `dataplaneapi_<v>_linux_<arch>.tar.gz` to `/usr/bin/dataplaneapi`. The tarball is verified against the per-release `checksums.txt` (anchored on the exact filename, tolerates `sha256sum` binary-mode `*` prefix).
- Ships `dataplaneapi.service` modeled on the upstream Debian package, with sandboxing added. Runs as `root` because the default `reload_strategy: systemd` invokes `systemctl reload haproxy.service`; docs explain how to drop-in a lower-privilege user with a socket-based reload.
- `tmpfiles.d` seeds `/etc/dataplaneapi/dataplaneapi.yml` (mode `0600 root:root`) from `/usr/share/dataplaneapi/dataplaneapi.yml` on first boot via `C` semantics — later sysext updates never overwrite a user-provided file. The shipped default targets the paths used by the `haproxy` sysext (`/usr/bin/haproxy`, `/etc/haproxy/haproxy.cfg`).
- Creates state dirs under `/var/lib/dataplaneapi/{transactions,backups}`.
- Wires `dataplaneapi latest` into `release_build_versions.txt` and adds `docs/dataplaneapi.md` plus an index row.

## Test plan

- [ ] `./bakery.sh list dataplaneapi` lists upstream releases
- [ ] `./bakery.sh create dataplaneapi vX.Y.Z` produces a working `dataplaneapi.raw` for x86-64 and arm64
- [ ] Booted alongside the `haproxy` sysext with a `userlist dataplaneapi` block in `haproxy.cfg`, `curl -u admin:... http://localhost:5555/v3/info` returns the running HAProxy version

### DIFF
--- a/docs/dataplaneapi.md
+++ b/docs/dataplaneapi.md
@@ -76,11 +76,17 @@ file at `/run/reboot-required`.
 You can deactivate updates by changing `enabled: true` to
 `enabled: false` in `systemd-sysupdate.timer`.
 
-Note that the snippet is for the x86-64 version of Data Plane API
-v3.3.5.
+The snippet bundles the [`haproxy` sysext](haproxy.md) alongside
+`dataplaneapi` because the API is only useful against a running HAProxy
+instance. It ships an `haproxy.cfg` with the `userlist dataplaneapi`
+block the API needs to accept requests.
 
-Check out the metadata release at
-https://github.com/flatcar/sysext-bakery/releases/tag/dataplaneapi for a
+Note that the snippet is for the x86-64 version of Data Plane API
+v3.3.5 and HAProxy 3.2.5.
+
+Check out the metadata releases at
+https://github.com/flatcar/sysext-bakery/releases/tag/dataplaneapi and
+https://github.com/flatcar/sysext-bakery/releases/tag/haproxy for a
 list of all versions available in the bakery.
 
 ```yaml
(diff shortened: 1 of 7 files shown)


t-lo:
@ananthb Looks good overall - sysext builds and `dataplaneapi --help` works for both amd64 and arm64. Pending changes in https://github.com/flatcar/sysext-bakery/pull/228#pullrequestreview-4617984566 so we can test it; if everything works we're good to merge!

docs/dataplaneapi.md
```diff
@@ -0,0 +1,121 @@
...
+    - path: /etc/sysupdate.dataplaneapi.d/dataplaneapi.conf
+      contents:
+        source: https://extensions.flatcar.org/extensions/dataplaneapi.conf
+    - path: /etc/sysupdate.d/noop.conf
+      contents:
+        source: https://extensions.flatcar.org/extensions/noop.conf
+  links:
+    - target: /opt/extensions/dataplaneapi/dataplaneapi-v3.3.5-x86-64.raw
+      path: /etc/extensions/dataplaneapi.raw
+      hard: false
```
t-lo, line 144: Let's add `haproxy.raw` and `haproxy.conf` from the bakery here too; standalone, the dataplaneapi sysext won't make much sense.

docs/dataplaneapi.md
```diff
@@ -0,0 +1,121 @@
...
+      enabled: true
+    - name: systemd-sysupdate.service
+      dropins:
+        - name: dataplaneapi.conf
+          contents: |
+            [Service]
+            ExecStartPre=/usr/bin/sh -c "readlink --canonicalize /etc/extensions/dataplaneapi.raw > /run/dataplaneapi-sysext"
+            ExecStartPre=/usr/lib/systemd/systemd-sysupdate -C dataplaneapi update
+            ExecStartPost=/usr/bin/sh -c "readlink --canonicalize /etc/extensions/dataplaneapi.raw > /run/dataplaneapi-sysext-new"
+            ExecStartPost=/usr/bin/sh -c "if ! cmp --silent /run/dataplaneapi-sysext /run/dataplaneapi-sysext-new; then touch /run/reboot-required; fi"
```
t-lo, line 164: Same as above, let's add a haproxy drop-in for systemd-sysupdate.

t-lo (changes requested):
A minor nit, documentation update only.

t-lo:
This looks great overall, and I successfully tested it with https://github.com/flatcar/sysext-bakery/pull/228. One more docs change and we can get it in.

t-lo (approved):
LGTM, thank you!

PR #233, opened 2026-07-08
vault: add HashiCorp Vault sysext

ananthb:
## Summary
- Adds a `vault` sysext that ships the OSS Vault release binary from `releases.hashicorp.com`, matching the layout of the existing `consul` and `nomad` sysexts.
- Bundles a `vault.service` systemd unit (Type=notify, `CAP_IPC_LOCK` + `LimitMEMLOCK=infinity` for Vault's mlock requirement, `UpheldBy=multi-user.target`) and the standard `multi-user.target.upholds/vault.service` symlink.
- Ships a static `vault` system user via `sysusers.d/vault.conf` (home `/var/lib/vault`, `/sbin/nologin`) instead of relying on `DynamicUser`, matching the coder sysext precedent — Vault operators regularly need stable UID ownership on the storage dir for backup/restore, which `DynamicUser`'s per-boot allocation doesn't provide.
- Shipped default config runs a loopback TCP listener over `storage "file"` at `/var/lib/vault/data` (under the sysusers home), with the UI on. `tmpfiles.d/10-vault.conf` creates `/var/lib/vault` and `/var/lib/vault/data` owned by the `vault` user. Meant to be replaced via a custom Butane config for production.
- Registered in `release_build_versions.txt` (`vault 1.20.0` + `vault latest`) and in the `docs/index.md` extensions table.

## Test plan
- [ ] `./bakery.sh list vault` returns OSS versions from the HashiCorp releases API.
- [ ] `./bakery.sh create vault 1.20.0` produces `vault.raw` on x86-64.
- [ ] `./bakery.sh create vault 1.20.0 --arch arm64` produces the arm64 sysext.
- [ ] `./bakery.sh boot vault` — after merge, `vault --version` reports the pinned version, `systemctl status vault` shows the service running, and the sysusers/tmpfiles entries create `/var/lib/vault` and materialize `/etc/vault.d/vault.hcl`.

### DIFF
--- a/vault.sysext/create.sh
+++ b/vault.sysext/create.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+# vim: et ts=2 syn=bash
+#
+
+RELOAD_SERVICES_ON_MERGE="true"
+
+# The HashiCorp releases API caps ?limit at 20 and returns newest first,
+# so walk pages via ?after=<timestamp> until a short page comes back
+# (last page) or the safety cap trips. Emit versions as they arrive so
+# `--latest true` can abort after the first fetch.
+function list_available_versions() {
+  local base="https://api.releases.hashicorp.com/v1/releases/vault"
+  local after=""
+  local max_pages=20
+  local curl_opts=(
+    -fsSL
+    --retry-delay 1 --retry 60 --retry-connrefused
+    --retry-max-time 60 --connect-timeout 20
+  )
+
+  local page
+  for page in $(seq 1 "$max_pages") ; do
+    local response
+    if [[ -z "$after" ]] ; then
+      response="$(curl "${curl_opts[@]}" "${base}?limit=20")"
+    else
+      response="$(curl "${curl_opts[@]}" -G \
+        --data-urlencode "limit=20" \
+        --data-urlencode "after=${after}" \
+        "${base}")"
+    fi
+
+    local count
+    count="$(printf '%s' "$response" | jq 'length')"
+    [[ "$count" -eq 0 ]] && break
+
+    printf '%s' "$response" | jq -r '.[] | select(.is_prerelease == false)
+                                         | select(.license_class == "oss")
+                                         | .version'
+
+    [[ "$count" -lt 20 ]] && break
+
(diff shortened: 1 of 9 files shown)


vault.sysext/create.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+# vim: et ts=2 syn=bash
+#
+
+RELOAD_SERVICES_ON_MERGE="true"
+
+function list_available_versions() {
+  curl -fsSL --retry-delay 1 --retry 60 \
+    --retry-connrefused --retry-max-time 60 --connect-timeout 20 \
+    https://api.releases.hashicorp.com/v1/releases/vault?limit=20 \
```
t-lo: Unfortunately this only returns the latest release, likely because of `limit=20`. Since the Hashicorp API does not allow for higher limits, we will need to handle pagination somehow, and process release information chunk by chunk using `after=` (see https://releases.hashicorp.com/docs/api/v1#operation/listReleasesV1).

`curl_api_wrapper` does something like that for GitHub releases and might serve as inspiration: https://github.com/flatcar/sysext-bakery/blob/main/lib/helpers.sh#L90

t-lo (changes requested):

vault.sysext/create.sh
```diff
@@ -0,0 +1,64 @@
...
+
+      printf '%s' "$response" | jq -r '.[] | select(.is_prerelease == false)
+                                           | select(.license_class == "oss")
+                                           | .version'
+
+      [[ "$count" -lt 20 ]] && break
+
+      after="$(printf '%s' "$response" | jq -r '.[-1].timestamp_created')"
+    done
+  } | sort -Vr
```
t-lo: `sort` is not needed as the HashiCorp API is guaranteed to return the latest release first, then go backwards in time. We can just print the versions as we receive them - this speeds up `./bakery.sh list vault --latest true` significantly since it can abort after the first fetch.

t-lo:
@ananthb Great progress - one minor nit and we're there!

t-lo (approved):
LGTM, tested for AMR64 and AMD64. Thank you for the good work!

PR #234, opened 2026-07-08
bcachefs-tools: add sysext for bcachefs userspace utilities

ananthb:
## Summary
- Adds a `bcachefs-tools` sysext shipping the [bcachefs userspace utilities](https://github.com/koverstreet/bcachefs-tools): the unified `bcachefs` CLI and its argv[0] dispatchers (`mkfs.bcachefs`, `fsck.bcachefs`, `mount.bcachefs`, `dump.bcachefs`) — so `mount -t bcachefs` picks up `mount.bcachefs` automatically once the sysext is merged.
- Statically builds bcachefs-tools inside an ephemeral Alpine container (Rust + C, Make-driven) with `+crt-static` for the Rust side and `PKG_CONFIG_ALL_STATIC=1` for the C side — same pattern as `scx.sysext`. This avoids any coupling to Flatcar's glibc.
- Redirects install targets that would land under `/usr/sbin` to `/usr/bin` (Flatcar's `/usr/sbin` is a symlink to `/usr/bin`; shipping the former in a sysext would clobber the symlink on merge). Belt-and-suspenders: if any binary still lands under `/usr/sbin`, it's relocated before packaging.
- Drops dev/man/pkgconfig files from the sysext — pure noise for a runtime sysext.
- Version listing uses `list_github_tags koverstreet/bcachefs-tools` since upstream tags releases but does not publish GitHub Releases.
- Registered in `release_build_versions.txt` (`bcachefs-tools latest`) and in the `docs/index.md` extensions table.

**Kernel module is out of scope for this PR.** The bcachefs kernel module is shipped separately by a companion `bcachefs-kmod` sysext (PR to follow) because the two have different lifecycles: userspace tools track bcachefs-tools upstream, while the kernel module is coupled to a specific Flatcar release's kernel and cannot share a version axis.

## Test plan
- [ ] `./bakery.sh list bcachefs-tools` returns upstream tags.
- [ ] `./bakery.sh create bcachefs-tools v1.25.2` produces `bcachefs-tools.raw` for x86-64.
- [ ] `./bakery.sh create bcachefs-tools v1.25.2 --arch arm64` produces the arm64 sysext.
- [ ] After merge on a Flatcar release whose kernel enables `CONFIG_BCACHEFS_FS`: `bcachefs version` reports the pinned version, `modprobe bcachefs && grep bcachefs /proc/filesystems` succeeds, and `mkfs.bcachefs` + `mount -t bcachefs` round-trip on a loopback device.
- [ ] On a Flatcar release without in-tree bcachefs: `bcachefs version` still works; kernel-side probing needs the `bcachefs-kmod` sysext (separate PR).

### DIFF
--- a/bcachefs-tools.sysext/build.sh
+++ b/bcachefs-tools.sysext/build.sh
@@ -1,0 +1,96 @@
+#!/bin/ash
+#
+# Build script helper for the bcachefs-tools sysext.
+# Runs inside an ephemeral Alpine container.
+# Builds the bcachefs userspace utilities and exports them to a
+# bind-mounted volume at /install_root.
+#
+set -euo pipefail
+
+version="$1"
+export_user_group="$2"
+
+# Build deps. bcachefs-tools is a Rust+C project orchestrated by Make,
+# linking against a fair pile of system libraries (sodium, urcu, blkid,
+# keyutils, zlib, zstd, lz4, aio, udev, attr).
+#
+# We try a static build first (matches the scx.sysext precedent). If a
+# given dep does not ship a static archive on Alpine, the corresponding
+# *-static package add will fail loudly during CI and we can swap to a
+# dynamic build that bundles the .so's under /usr/lib64.
+apk add --no-cache \
+  clang \
+  clang-dev \
+  clang-libclang \
+  llvm \
+  lld \
+  bpftool \
+  libbpf-dev \
+  elfutils-dev \
+  pkgconf \
+  curl \
+  git \
+  make \
+  gcc \
+  musl-dev \
+  linux-headers \
+  ca-certificates \
+  cargo \
+  rust \
+  zlib-dev          zlib-static \
+  zstd-dev          zstd-static \
+  lz4-dev           lz4-static \
+  libsodium-dev     libsodium-static \
+  liburcu-dev \
+  util-linux-dev \
+  keyutils-dev \
+  libaio-dev \
+  eudev-dev \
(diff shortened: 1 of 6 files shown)


ananthb:
Closing this too — after digging into how ZFS is packaged on Flatcar, the honest answer is that both bcachefs sysexts should live inside `flatcar/scripts` as a proper Flatcar OS extension, not in the bakery.

Concretely: Gentoo's `sys-fs/bcachefs-tools` ebuild (already in the main tree) has `MODULES_OPTIONAL_IUSE=+modules` and inherits `linux-mod-r1`, so a single package builds the userspace tools **and** the kernel module together, against the exact kernel Flatcar's SDK is building. That means one Flatcar-native `flatcar-bcachefs` sysext replaces both this PR and #235:
- Kernel binding for free (built against `sys-kernel/coreos-modules`).
- Ephemeral per-release signing for secure boot.
- Native update flow (users write `bcachefs` into `/etc/flatcar/enabled-sysext.conf`; Flatcar's own updater pulls the matching version at every OS upgrade).
- Follows the exact same pattern as the existing ZFS sysext (also an out-of-tree module, similar stability caveats).

A bakery-side tools-only sysext would provide no standalone value on its own — bcachefs userspace with no kernel module can't mount or manage anything, so it's dead weight unless paired with a module sysext (which we've now decided lives on the OS side, not the bakery side).

New home for both: **https://github.com/flatcar/scripts** — PR incoming.

PR #235, opened 2026-07-08
bcachefs-kmod: add sysext shipping a prebuilt bcachefs kernel module

ananthb:
## Summary
Companion to #234 (the \`bcachefs-tools\` sysext) — this PR adds \`bcachefs-kmod\`, which ships a prebuilt \`bcachefs.ko\` for a specific Flatcar release's kernel.

- Version listing tracks \`koverstreet/bcachefs-tools\` tags — the kernel-side bcachefs source is versioned alongside userspace tools.
- Optional \`--flatcar-release <version|channel>\` parameter (default \`stable\`) picks which Flatcar release's kernel to build against. Channel names (\`alpha\`, \`beta\`, \`stable\`, \`lts\`) resolve to the current release on that channel.
- \`build.sh\` runs inside an ephemeral Alpine container and:
  1. Resolves the Flatcar release → kernel version via \`flatcar/scripts@<channel>-<release>\`'s kernel Manifest.
  2. Fetches vanilla \`linux-<kver>.tar.xz\` from \`cdn.kernel.org\` and applies Flatcar's kernel config.
  3. Overlays \`fs/bcachefs/\` from \`koverstreet/bcachefs\` at the requested tag onto the vanilla tree.
  4. Builds \`bcachefs.ko\` out-of-tree via \`make M=fs/bcachefs modules\`.
  5. Installs to \`/usr/lib/modules/<kver>/updates/bcachefs.ko\` with \`depmod\` metadata, plus a \`modules-load.d/bcachefs.conf\` for auto-load on boot.

## Deliberately not in the automated release matrix
A single build pins **two** dimensions (bcachefs revision × Flatcar release/kernel), so:
- Not added to \`release_build_versions.txt\` — the \`latest\` scanner would produce a single output per bcachefs tag, which doesn't map to the (bcachefs × Flatcar release) matrix this sysext lives in.
- Not eligible for \`systemd-sysupdate\` auto-updates — the \`.ko\` will only load on the exact matching kernel, so silently rolling forward on a kernel bump would break \`modprobe\`.
- Sysext filename encodes both axes: \`bcachefs-kmod-<bcachefs-tag>-flatcar-<release>-<arch>.raw\`.

Users rebuild against the new release on every Flatcar upgrade; \`docs/bcachefs-kmod.md\` calls this out.

## Stacking / merge order
This branch is stacked on \`add-bcachefs-tools-sysext\` (#234) because \`docs/bcachefs-kmod.md\` links to \`bcachefs-tools.md\` and the two sysexts travel together semantically. Please merge #234 first; happy to rebase this onto \`main\` afterwards if needed.

## Known caveats worth flagging in review
- Flatcar's kernel config path in \`flatcar/scripts\` has shifted across releases; \`build.sh\` tries a few well-known locations, but the first CI runs may need path adjustments. The candidate list is centralised at the top of \`build.sh\` for easy tweaks.
- The out-of-tree overlay approach assumes bcachefs source at the requested tag compiles against the vanilla kernel version Flatcar ships. If bcachefs upstream has drifted past that kernel, users need to pick a bcachefs tag that was current when the Flatcar release was cut.

## Test plan
- [ ] \`./bakery.sh list bcachefs-kmod\` returns bcachefs-tools upstream tags.
- [ ] \`./bakery.sh create bcachefs-kmod v1.25.2 --flatcar-release stable\` produces a sysext.
- [ ] \`./bakery.sh create bcachefs-kmod v1.25.2 --flatcar-release lts\` produces a sysext.
- [ ] \`./bakery.sh create bcachefs-kmod v1.25.2 --flatcar-release stable --arch arm64\` produces an arm64 sysext.
- [ ] On a Flatcar release that matches the \`--flatcar-release\` used: \`modprobe bcachefs && grep bcachefs /proc/filesystems\` succeeds after merge; loading fails with \`Exec format error\` on a mismatched Flatcar release, and the doc guides the user to rebuild.

### DIFF
--- a/bcachefs-kmod.sysext/build.sh
+++ b/bcachefs-kmod.sysext/build.sh
@@ -1,0 +1,95 @@
+#!/bin/bash
+#
+# Build script helper for the bcachefs-kmod sysext.
+# Runs inside the Flatcar SDK container (ghcr.io/flatcar/flatcar-sdk-all).
+# Builds bcachefs.ko using koverstreet/bcachefs-tools's dkms/ harness
+# against the kernel source shipped by the SDK, then exports the .ko
+# plus a modules-load.d snippet to the bind-mounted /install_root.
+set -euo pipefail
+
+bcachefs_version="$1"     # koverstreet/bcachefs-tools tag, e.g. v1.25.2
+export_user_group="$2"
+
+# 1. Fetch Flatcar's Portage tree + kernel sources ------------------------
+#
+# The SDK image ships emerge but not the Portage tree; emerge-gitclone
+# clones it on first use. coreos-modules pulls in coreos-sources as a
+# dependency, drops the kernel source with Flatcar's patches applied at
+# /usr/src/linux-<kver>-flatcar, and runs modules_prepare so we end up
+# with a build-usable KDIR (Module.symvers, generated headers, etc.).
+emerge-gitclone
+emerge -gKv sys-kernel/coreos-modules
+
+# 2. Resolve KDIR + kernel release string ---------------------------------
+kdir="$(readlink -f /usr/src/linux)"
+kver="$(basename "${kdir}")"
+kver="${kver#linux-}"    # e.g. "6.6.86-flatcar"
+
+echo "==> Building bcachefs ${bcachefs_version} against kernel ${kver}"
+
+# 3. Clone bcachefs-tools at the requested tag ----------------------------
+#
+# The kernel-side bcachefs source and its out-of-tree Kbuild harness live
+# in koverstreet/bcachefs-tools now — everything since the 6.18 mainline
+# removal ships through this repo (see dkms/ subdirectory).
+cd /tmp
+git clone --depth 1 --branch "${bcachefs_version}" --single-branch \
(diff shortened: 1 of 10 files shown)


ananthb:
Closing this — moving the kernel-module build out to an external repo. It's a better fit there:

- The two-axis versioning (bcachefs revision × Flatcar release/kernel) opts out of everything sysext-bakery specifically automates (single-axis `release_build_versions.txt` polling, hosted releases per-version, `systemd-sysupdate` auto-update).
- Every Flatcar SDK bump can break the build for reasons unrelated to any other sysext here, which would put maintenance load on this repo's maintainers for something they'd otherwise get no signal about.
- Kent Overstreet's blessed downstream path since the 6.18 removal is DKMS, so users who want bcachefs on Flatcar are already outside the "just merge a sysext" story — an external repo with clear build instructions honors that.

The companion `bcachefs-tools` sysext (#234) stays here — it fits the automation model perfectly and there's no reason to split it out.

New home for the kmod build: **https://github.com/ananthb/flatcar-bcachefs-kmod**.

PR #239, opened 2026-07-20
cni-plugins: new sysext for the containernetworking reference plugins

ananthb:
Adds a `cni-plugins` sysext shipping the upstream [containernetworking/plugins](https://github.com/containernetworking/plugins) reference CNI plugins under `/opt/cni/bin`.

## Why

`/opt/cni/bin` is the de-facto default CNI plugin directory. Nomad's `bridge` network
mode (and dynamic container ports / Consul Connect) work.

## What

- `cni-plugins.sysext/create.sh` — downloads the release tarball for the target arch, verifies the upstream-published `.sha256`, extracts into `/opt/cni/bin`.
- `docs/cni-plugins.md` — usage + butane snippet.
- `release_build_versions.txt` — `cni-plugins latest`.

Verified on Flatcar stable that plugins at the CNI path make Nomad bridge networking + dynamic ports work.

### DIFF
--- a/cni-plugins.sysext/create.sh
+++ b/cni-plugins.sysext/create.sh
@@ -1,0 +1,41 @@
+#!/usr/bin/env bash
+# vim: et ts=2 syn=bash
+#
+# CNI reference plugins system extension.
+#
+# Ships the upstream containernetworking/plugins binaries (bridge, host-local,
+# portmap, firewall, loopback, macvlan, ptp, …) under /opt/cni/bin — the
+# default plugin directory searched by Nomad (cni_path) and usable by any CNI
+# runtime (Consul Connect, container runtimes, …). There is no service; the
+# binaries are invoked by the CNI consumer.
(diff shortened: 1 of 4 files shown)


t-lo:
@ananthb Looks good in general, but extending `/opt` can cause unintended side effects as many Flatcar users assume it to be writable. I understand that's the default location but I'd prefer if we could ship the binaries in `/usr` instead, as it would integrate much more cleanly with Flatcar (where `/usr` is already read-only, so no expectations are violated).

I wonder if it's even useful to ship CNI separately or if we should update sysexts of components that need it, and provide shipping them in a library these sysexts import. The `nerdctl` sysext for instance implements optional shipping of CNI.

ananthb:
The nerdctl approach is interesting.

ananthb:
@t-lo I've switched to optionally shipping cni plugins in nomad.

PR #240, opened 2026-07-20
dataplaneapi: reload HAProxy without sudo (root unit)

ananthb:
The dataplaneapi sysext ships a `dataplaneapi.yml` with
`reload_strategy: systemd`, which makes the Data Plane API shell out to
`sudo -n systemctl reload haproxy` on every config change.

But the shipped `dataplaneapi.service` runs the process as **root**, so:

- `sudo` is unnecessary, and
- `sudo -n` **fails** on hosts without a passwordless sudoers entry for root
  (e.g. Flatcar), with `exit status 1`.

A failed reload is worse than a no-op: the Data Plane API **reverts its last
working config**. In practice, e.g. Consul service-discovery backends get
created in the API model but are never applied to HAProxy — they silently
disappear, which is very hard to diagnose.

This switches the shipped config to a `custom` reload strategy that calls
`systemctl` directly (`reload_cmd`/`restart_cmd`), which works for the root
unit. Verified on Flatcar stable: with `systemd`+sudo the reloads fail and
Consul-discovered backends never load; with `custom` they apply correctly.
Also updates the Privileges section of the docs.

### DIFF
--- a/dataplaneapi.sysext/files/usr/lib/systemd/system/dataplaneapi.service
+++ b/dataplaneapi.sysext/files/usr/lib/systemd/system/dataplaneapi.service
@@ -17,7 +17,7 @@ Restart=on-failure
 RestartSec=5
 
 # Dataplane API needs to invoke `systemctl reload haproxy.service`
-# (reload_strategy: systemd in dataplaneapi.yml), so it runs as root.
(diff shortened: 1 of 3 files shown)


t-lo (approved):
LGTM, thanks!

PR #241, opened 2026-07-22
Add node_exporter as sysext

ananthb:
This adds the [Prometheus node_exporter](https://github.com/prometheus/node_exporter) as a system extension.

## What's included
- `node_exporter.sysext/create.sh`: lists releases from `prometheus/node_exporter`, downloads the release tarball for the target arch, verifies it against the upstream `sha256sums.txt`, and installs the `node_exporter` binary to `/usr/bin`.
- A `node_exporter.service` unit that runs the exporter as a `DynamicUser` with hardening (`ProtectSystem=strict`, `NoNewPrivileges`, restricted address families). It is started on merge via a `multi-user.target.upholds` drop-in.
- A default `/etc/node_exporter/node_exporter.env` (seeded via tmpfiles) for passing extra flags through `NODE_EXPORTER_OPTS`.
- Docs page and entries in `docs/index.md` and `release_build_versions.txt` (`v1.12.1` + `latest`).

## Testing
- `./bakery.sh list node_exporter` returns the upstream version list.
- `./bakery.sh create node_exporter v1.12.1 --arch x86-64` downloads and passes checksum verification.


ananthb:
Closing in favour of #255

PR #242, opened 2026-07-22
Add blackbox_exporter as sysext

ananthb:
This adds the [Prometheus blackbox_exporter](https://github.com/prometheus/blackbox_exporter) as a system extension.

## What's included
- `blackbox_exporter.sysext/create.sh`: lists releases from `prometheus/blackbox_exporter`, downloads the release tarball for the target arch, verifies it against the upstream `sha256sums.txt`, and installs the `blackbox_exporter` binary to `/usr/bin` plus the upstream default `blackbox.yml` to `/usr/share/blackbox_exporter`.
- A `blackbox_exporter.service` unit that runs as a `DynamicUser` with `AmbientCapabilities=CAP_NET_RAW` (needed for the ICMP prober) and hardening. It reads its config from `/etc/blackbox_exporter/blackbox.yml` and is started on merge via a `multi-user.target.upholds` drop-in.
- The upstream default probe config and an environment file are seeded into `/etc/blackbox_exporter/` via tmpfiles; extra flags can be passed through `BLACKBOX_EXPORTER_OPTS`.
- Docs page and entries in `docs/index.md` and `release_build_versions.txt` (`v0.28.0` + `latest`).

## Testing
- `./bakery.sh list blackbox_exporter` returns the upstream version list.
- `./bakery.sh create blackbox_exporter v0.28.0 --arch x86-64` downloads and passes checksum verification; the tarball contains both the binary and `blackbox.yml` as expected.

### DIFF
--- a/blackbox_exporter.sysext/create.sh
+++ b/blackbox_exporter.sysext/create.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bash
+# vim: et ts=2 syn=bash
+#
+# Prometheus blackbox_exporter sysext.
+#
+
+RELOAD_SERVICES_ON_MERGE="true"
+
+function list_available_versions() {
+  list_github_releases "prometheus" "blackbox_exporter"
+}
+# --
+
+function populate_sysext_root() {
+  local sysextroot="$1"
+  local arch="$2"
+  local version="$3"
+
+  local rel_arch="$(arch_transform "x86-64" "amd64" "$arch")"
+  local tarball="blackbox_exporter-${version#v}.linux-${rel_arch}.tar.gz"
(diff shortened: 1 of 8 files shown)


t-lo:
@ananthb Is there a good reason for shipping this as a sysext in contrast to shipping it in a container? I feel the whole prometheus / blackbox toolset has solid upstream and community support when shipped as containers (or pods, for Kubernetes) - there is a lot of prior art around that users can benefit from (processes, example configs, etc.). What's the value of having it in a sysext?

ananthb:
Yeah I can close this and run it as a container. I also see sysext-bakery as absolute essentials so I see where you're coming from.

PR #245, opened 2026-08-04
Add qemu as sysext (full-VM hypervisor)

ananthb:
Adds a `qemu` sysext shipping `qemu-system-x86_64`, e.g. as the backing hypervisor for Nomad's `qemu` task driver on Flatcar.

QEMU has no upstream static/portable release, so the recipe bundles the binary with its full shared-library closure (glibc + dynamic loader) and firmware (SeaBIOS, iPXE option ROMs, OVMF), extracted from a Debian container. `tools/flix.sh` resolves the closure and patchelf's the binary onto a private loader and rpath, isolating it from the host's libraries (same approach as `tilde.sysext`). No wrapper script and no `-L` flag are needed: QEMU finds the bundled firmware through its compiled-in data directory (`/usr/share/qemu`), which the sysext ships. User-mode (SLIRP) networking is included so guests get NAT egress with no host bridge setup.

- `qemu.sysext/create.sh`: resolve the Debian suite that ships the requested QEMU version, extract, and bundle via `tools/flix.sh`
- `docs/qemu.md` + index entry
- `release_build_versions.txt`: `qemu 10.0.11`

The version parameter is a QEMU release (`x.y.z` or `latest`); the recipe resolves whichever Debian suite currently ships it and verifies the installed binary matches, so a moved suite cannot silently produce a different build.

x86-64 only for now. Verified on a Flatcar node: the merged sysext's `qemu-system-x86_64` runs and `-netdev user` (SLIRP) works, booting a NixOS guest that reaches multi-user with NAT egress.

Note: this is an unconventional recipe (distro-extraction + glibc bundle) since there's no upstream portable qemu — open to a cleaner approach if maintainers prefer.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -86,6 +86,7 @@ Check out documentation on specific extensions at the navigation menu on the lef
 | `nvidia-runtime` |  released    | [nvidia-runtime versions](https://github.com/flatcar/sysext-bakery/releases/tag/nvidia-runtime) |
 | `ollama`         |  released    | [ollama versions](https://github.com/flatcar/sysext-bakery/releases/tag/ollama) |
 | `opkssh`         |  released    | [opkssh versions](https://github.com/flatcar/sysext-bakery/releases/tag/opkssh) |
+| `qemu`           |  released    | [qemu versions](https://github.com/flatcar/sysext-bakery/releases/tag/qemu) |
 | `rke2`           |  released    | [rke2 versions](https://github.com/flatcar/sysext-bakery/releases/tag/rke2) |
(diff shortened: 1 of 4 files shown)


pothos:
Interesting! Your setup somehow goes into the same direction as flix, I think you could use that instead. See https://github.com/flatcar/sysext-bakery/blob/main/tilde.sysext/create.sh for an example.

t-lo:
Hmmm, any chance to get you to statically compile it from sources in an ephemeral Alpine container? We use that approach for other sysexts and it would make us independent from Debian qemu releases and its library dependencies.

ananthb:
I'm willing to give that a go. I did take that approach on another one of my PRs. Happy to try it again.

PR #254, opened 2026-08-10
test: add automated sysext validation

ravindra-RKB:
## Summary

* add a `bakery.sh test` command for automated sysext validation
* validate the extension through a temporary Flatcar VM using the existing `boot_sysext` infrastructure
* add CI coverage for changed extensions on both x86-64 and arm64
* support extension-specific dynamic tests through `test.sh`
* document the new testing workflow

## Validation

* `bash -n bakery.sh`
* `bash -n lib/test.sh`
* `shellcheck bakery.sh`
* `shellcheck lib/test.sh`
* `yamllint .github/workflows/pull_request.yaml`
* `git diff --check`
* successfully validated the implementation against a real compiled sysext image

### DIFF
--- a/.github/workflows/pull_request.yaml
+++ b/.github/workflows/pull_request.yaml
@@ -1,0 +1,92 @@
+---
+name: Validate Sysext Changes
+
+"on": [pull_request]
+
+jobs:
+  dispatch:
+    runs-on: ubuntu-24.04
+    outputs:
+      tests: ${{ steps.filter.outputs.tests }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


t-lo:
Pardon for the lack of documentation, but `test` is envisioned to boot a Flatcar VM and run tests on a live system (e.g. checking if a systemd unit was started successfully, or if there were any errors during boot). The imperative nature of sysext build scripts makes it unnecessary to check whether individual files exist since we just put them there - so checking afterwards doesn't have much value except than proving that `cp` worked.

I'll close this but will be happy to discuss what `test` was originally planned to be on discord or Matrix.

PR #255, opened 2026-08-13
node-exporter: add prometheus node-exporter system extension

amanvaibhav162:
# node-exporter: add prometheus node-exporter system extension

This PR adds the Prometheus Node Exporter system extension (`node-exporter.sysext`) to the Bakery. The extension downloads the official upstream Node Exporter binary and packages it into a system extension image alongside a systemd unit file and default configuration.

The Node Exporter service is integrated using a systemd `Upholds` drop-in on `multi-user.target` to ensure the service automatically starts and remains active when the system enters multi-user mode. Default environment options are shipped under `/usr/share/node-exporter/node-exporter.env` and loaded by the service unit, which also supports user-provided overrides from `/etc/default/node-exporter`.

Additionally, user documentation has been added under `docs/node_exporter.md` with instructions on how to use, configure, and automate updates for this extension using Butane and `systemd-sysupdate`.

## How to use

Reviewers can build and run the Node Exporter system extension locally using the Bakery script:

1. Build the system extension image:
   ```bash
   ./bakery.sh create node-exporter v1.12.1
   ```

2. Boot a local Flatcar VM with the built system extension:
   ```bash
   ./bakery.sh boot node-exporter.raw --ports 9200:9100
   ```

3. Log in to the VM and verify the status of the service, and verify that the metrics endpoint is responding on the forwarded host port `9200`.

## Testing done

- Built the extension successfully using `./bakery.sh create node-exporter v1.12.1`.
- Booted the extension in QEMU via `./bakery.sh boot node-exporter.raw --ports 9200:9100`.
- Verified the binary is present and correct:
  ```bash
  core@localhost ~ $ node_exporter --version
  node_exporter, version 1.12.1 (branch: HEAD, revision: 6044da783597cc3b57aef7580ddcdcff58a4ee99)
    build user:       root@3a6b59999579
    build date:       20260714-12:05:11
    go version:       go1.26.5
    platform:         linux/amd64
    tags:             unknown
  ```
- Verified the systemd service starts automatically at boot and runs successfully:
  ```bash
  core@localhost ~ $ systemctl status node-exporter.service
  ● node-exporter.service - Prometheus Node Exporter
       Loaded: loaded (/usr/lib/systemd/system/node-exporter.service; disabled; preset: enabled)
       Active: active (running) since Thu 2026-08-13 18:09:39 UTC; 12min ago
     Invocation: 2aa630f4dfe74435b64ea922e66d696b
           Docs: https://github.com/prometheus/node_exporter
       Main PID: 2072 (node_exporter)
          Tasks: 7 (limit: 15176)
         Memory: 26.4M (peak: 27.5M)
            CPU: 89ms
         CGroup: /system.slice/node-exporter.service
                 └─2072 /usr/bin/node_exporter
  ```
- Verified metrics collection on the host machine via forwarded port `9200`:
  ```bash
  curl http://localhost:9200/metrics
  ```
  Output returned correct Prometheus Exposition Text Format metrics successfully (e.g. `go_gc_duration_seconds{quantile="0"} 1.9324e-05` and various host stats).

- [ ] Changelog entries added in the respective `changelog/` directory (Note: changelog/ directory is not used in this repository)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -81,6 +81,7 @@ Check out documentation on specific extensions at the navigation menu on the lef
 | `kubernetes`     |  released    | [kubernetes versions](https://github.com/flatcar/sysext-bakery/releases/tag/kubernetes) |
 | `llamaedge`      |  released    | [llamaedge versions](https://github.com/flatcar/sysext-bakery/releases/tag/llamaedge) |
 | `nebula`         |  released    | [nebula versions](https://github.com/flatcar/sysext-bakery/releases/tag/nebula) |
+| `node-exporter`  |  released    | [node-exporter versions](https://github.com/flatcar/sysext-bakery/releases/tag/node-exporter) |
 | `nerdctl`        |  released    | [nerdctl versions](https://github.com/flatcar/sysext-bakery/releases/tag/nerdctl) |
 | `nomad`          |  released    | [nomad versions](https://github.com/flatcar/sysext-bakery/releases/tag/nomad) |
 | `nvidia-runtime` |  released    | [nvidia-runtime versions](https://github.com/flatcar/sysext-bakery/releases/tag/nvidia-runtime) |
--- a/docs/node_exporter.md
+++ b/docs/node_exporter.md
@@ -1,0 +1,51 @@
+# Prometheus Node Exporter sysext
+
+This sysext ships [Prometheus Node Exporter](https://github.com/prometheus/node_exporter).
+
+The sysext includes a service unit file to start node-exporter at boot.
+The default configuration can be modified or replaced via a custom Butane config.
+
+# Usage
+
+The snippet includes automated updates via systemd-sysupdate.
+Sysupdate will stage updates, refresh the merged sysext, and restart `node-exporter.service` — no reboot is required.
+You can deactivate updates by changing `enabled: true` to `enabled: false` in `systemd-sysupdate.timer`.
+
+Note that the snippet is for the x86-64 version of node-exporter 1.12.1.
+
+Check out the metadata release at https://github.com/flatcar/sysext-bakery/releases/tag/node-exporter for a list of all versions available in the bakery.
+
+```yaml
+variant: flatcar
+version: 1.0.0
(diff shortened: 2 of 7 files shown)


amanvaibhav162:
Hi @tormath1,

Could you please take a look at this new system extension recipe? 

### Purpose
It adds support for **node-exporter** to provide a detailed system overview of the whole node.

### What to Check
* The recipe packaging logic in `create.sh`.
* The configuration and service files under `files/`.

Thanks for your time and feedback!

ananthb:
I've got #241 open that does the same thing already.

t-lo:
@amanvaibhav162  Is there a good reason for shipping this as a sysext in contrast to shipping it in a container? I feel the whole prometheus node exporter toolset has solid upstream and community support when shipped as containers (or pods, for Kubernetes) - there is a lot of prior art around that users can benefit from (processes, example configs, etc.). What's the value of having it in a sysext?

amanvaibhav162:
@t-lo Thank you Sir for reviewing, and the answer to your question,
**Runtime independence**: If the container runtime crashes or is updating then a containerized exporter dies with it. As a sysext, it runs as a native systemd unit with zero dependency on Docker/containerd, so you keep visibility exactly when it matters most.
**Boot-time metrics** :Sysexts merge early in boot, well before any container runtime starts. You get host metrics from first boot, not after the full container stack is up.
**No privileged container workarounds** :Node-exporter needs /proc, /sys, and host root access. In a container, that means --privileged, --pid=host, --net=host, and bind mounts — effectively host access with extra steps. A sysext avoids that entirely.
**Bare-metal & non-K8s use cases** : Not every Flatcar node runs Kubernetes. For standalone or edge deployments, a sysext gives monitoring out of the box without requiring a container runtime just for metrics.

This isn't meant to replace the containerized approach for Kubernetes environments (where DaemonSets and the Prometheus operator ecosystem are excellent). It covers the cases where that infrastructure isn't present or can't be relied upon.
Please review my PR#256 too

PR #257, opened 2026-08-17
Teleport

reavessm:
# Add Teleport sysext

Add the tctl, tsh, and teleport binaries via sysex

## How to use

```yaml
variant: flatcar
version: 1.0.0

storage:
  files:
    - path: /opt/extensions/teleport/teleport-v18.9.2-x86-64.raw
      mode: 0644
      contents:
        source: https://example.com/teleport-v18.9.2-x86-64.raw
    - path: /etc/teleport/teleport.yaml
      mode: 0640
      contents:
        source: data:text/plain;charset=utf-8,<url-encoded-config>
  links:
    - target: /opt/extensions/teleport/teleport-v18.9.2-x86-64.raw
      path: /etc/extensions/teleport.raw
      hard: false
    - path: /etc/systemd/system/multi-user.target.wants/teleport.service
      target: /usr/lib/systemd/system/teleport.service
      overwrite: true
```

## Testing done

Created the .raw file via `./bakery.sh create teleport v18.9.2 --sysupdate true`, mounted it, validated the binaries are there.


docs/teleport.md
```diff
@@ -0,0 +1,73 @@
...
+        - name: teleport.conf
+          contents: |
+            [Service]
+            ExecStartPre=/usr/bin/sh -c "readlink --canonicalize /etc/extensions/teleport.raw > /tmp/teleport"
+            ExecStartPre=/usr/lib/systemd/systemd-sysupdate -C teleport update
+            ExecStartPost=/usr/bin/sh -c "readlink --canonicalize /etc/extensions/teleport.raw > /tmp/teleport-new"
+            ExecStartPost=/usr/bin/sh -c "if ! cmp --silent /tmp/teleport /tmp/teleport-new; then systemd-sysext refresh && systemctl restart teleport.service; fi"
+```
+
+The `teleport.conf` sysupdate transfer config should contain:
```
tormath1: I think we can drop this section as the automation will generate the `teleport.conf` for us.

tormath1 (commented):
Thanks for your contribution, it looks good. Can you add `teleport latest` to the list of built sysext in ./release_build_versions.txt ?

reavessm:
Hey @tormath1, can I get another review?  Thank you

PR #262, opened 2026-08-27
Drop coder

ananthb:
# Drop coder

Coder is better off being run as a container than as a sysext. Nothing is gained from running it baremetal.


ananthb:
@t-lo in the same spirit as blackbox_exporter, it makes sense to drop coder as a sysext and run it as a container. Upstream recommends running coder as a container too.

